PR #2, opened 2020-01-24
docs around the JSON contract for skip payment

jrobinson01:
This is the best way I can think of the share the json contracts for power-ons. Git also gives us nice visibility around changes to them.


svetkomir:
Looks good!

PR #8, opened 2020-03-02
withdraw-by-check error docs

jrobinson01:
adds errors for "missing address" and "invalid loan/share"
@tkainz I left the logging message the same as the other errors but feel free to make suggestions. I believe these errors can occur in both the `STATESTART` and `PERFORMWITHDRAW` states.

### DIFF
--- a/withdraw-by-check/withdraw-by-check-json-contract.md
+++ b/withdraw-by-check/withdraw-by-check-json-contract.md
@@ -108,6 +108,24 @@ If the request is not successful due to insufficient funds, the poweron should r
 }
 ```
 
+If the request is not successful due to a missing address, the poweron should respond with:
+
+```json
+{
+  "errorCode": 502,
+  "loggingErrorMessage": "TRANPERFORM Error:+TRANERROR"
+}
+```
+
+If the request is not successful due to an invalid loan/share, the poweron should respond with:
+
+```json
+{
+  "errorCode": 503,
+  "loggingErrorMessage": "TRANPERFORM Error:+TRANERROR"
+}
+```
+
 In the case of insufficient funds, the client may try request again with a lesser amount.
 
 ### Error codes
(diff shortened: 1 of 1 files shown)


tkainz:
@jrobinson01  - Currently have the program coded (logging message) for the following error conditions:
"Account Warning Found"
"Invalid Share Type"
"Invalid Loan Type"
"Share Warning Found"
"Loan Warning Found"
"Invalid Address"
And the actual error thrown by the attempted open or read of the config file.

jrobinson01:
Looks good to me @tkainz . You're clear on what error codes to map those to?

tkainz:
@jrobinson01 - Just following the Yellow Brick.....I mean the JSON Contract!

withdraw-by-check/withdraw-by-check-json-contract.md
```diff
@@ -108,10 +108,30 @@ If the request is not successful due to insufficient funds, the poweron should r
 }
 ```
 
+If the request is not successful due to a missing address, the poweron should respond with:
+
+```json
+{
+  "errorCode": 502,
+  "loggingErrorMessage": "TRANPERFORM Error:+TRANERROR"
```
tkainz: Tranperform error?

jrobinson01: I think that's what Christina originally sent over. I'm going to leave it to you to fill that in with whatever makes sense. We can do something simple like use the strings you mentioned https://github.com/Banno/banno-powerons/pull/8#issuecomment-593438132 or add to them if there's anything else you think would be useful.

withdraw-by-check/withdraw-by-check-json-contract.md
```diff
@@ -108,10 +108,30 @@ If the request is not successful due to insufficient funds, the poweron should r
...
+}
```
tkainz: How is this error code situation (503) different than the error code situation described on line 55?

jrobinson01: Good catch. Looks like a duplicate. I'll adjust.

PR #9, opened 2020-03-04
Updating files with current versions

tkainz:
Added current version of Skip Payment and Loan Payoff PowerOns and associated CFG Letter files


christinasimms:
I have reviewed the changes and they are correct.

PR #19, opened 2020-04-01
Update BANNO.CHECK.WITHDRAW.V1.POW

tkainz:
Added check and processing of WD amounts w or w/o decimal. Added address validation. Added memomode check. Some formatting and commenting.


christinasimms:
I am sure you tested this via SoapUI.  Just want to verify you tested by logging into Banno online and mobile.

PR #20, opened 2020-04-06
initial open-subshare contract

jrobinson01:


open-subshare/open-subshare-json-contract.md
```diff
@@ -0,0 +1,128 @@
...
+```
+
+To which the poweron should respond with:
+NOTE: if memoMode true, categories won't exist
+
+```json
+{
+  "memoMode": false,
+  "results": {
+    "canOpenSubShare": true,
```
tkainz, line 31: @jrobinson01 , Would the 'canOpenSubshare' element be returned by the PowerOn if there were an error condition being returned as well... as would be the case if there were no eligible shares which could be created?

{
	"memoMode": false,
	"results": {
		"canOpenSubShare": true,
		"errorCode": 503,
		"loggingErrorMessage": "No eligible shares available to be created"
	}
}

open-subshare/open-subshare-json-contract.md
```diff
@@ -0,0 +1,128 @@
...
+
+NOTE: the 'electronic' type should not be present if the member does not have any valid shares to fund with.
+
+```json
+{
+  "memoMode": false,
+  "results": {
+    "categoryTerms":["array ", "of ", "strings."],
+    "fundingOptions":["electronic", "check", "later"],
+    "electronicFundingAccounts":[
```
tkainz, line 82: Should we also be returning the minimum funding requirement here as well?

    "categoryTerms":["array ", "of ", "strings."],
    **"fundingRequirement": "         100.00",**
    "fundingOptions":["electronic", "check", "later"],
    "electronicFundingAccounts":[

open-subshare/open-subshare-json-contract.md
jrobinson01: I think we can stick to the format we came up with for wd-by-check. In this case:
```json
{
  "errorCode": 503,
  "loggingErrorMessage": "TRANPERFORM Error:+TRANERROR"
}
```
I'll start an Errors section in the doc

open-subshare/open-subshare-json-contract.md
```diff
@@ -30,14 +30,32 @@ Poweron response:
...
+            "shareType": "0",
             "name": "7 Month certificate",
             "term": "7 months",
             "minBalance": "1000.00",
-            "interestRate": "2.450%" // optional
+            "interestRates": [
+              {
+                "rate": "0.000",
+                "minBalance": "0.00",
+                "maxBalance": "4.99",
```
tkainz, line 44: I think the comma after the "maxBalance" values isn't needed because that's the last element in the group

PR #21, opened 2020-04-09
Update BANNO.CHECK.WITHDRAW.V1.POW

tkainz:
Updated to allow specific characters ony in TOC and Share/Loan Description and to check for back-slash and double-quotes an if found, prepend character with a black-slash

### DIFF
--- a/withdraw-by-check/REPWRITERSPECS/BANNO.CHECK.WITHDRAW.V1.POW
+++ b/withdraw-by-check/REPWRITERSPECS/BANNO.CHECK.WITHDRAW.V1.POW
@@ -4,6 +4,8 @@
 **
 **  Copyright 2020 Jack Henry and Associates
 **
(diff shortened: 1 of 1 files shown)


christinasimms:
I am sure you tested this via SOAPUI, but was it tested by logging into Banno online and mobile?

tkainz:
Tested in Banno but not yet in mobile..... still working on trying to get the apk sideloaded on my phone

tkainz:
Ok... tested via Soap, Banno and Mobile!

christinasimms:
Since there are versions of this that are live, can you please add a note at the top like this:

**  4/16/2020:  fixes for invalid punctuation in share descriptions

In Episys, we cannot see the Githib versions and this will help support.  Once a new version (ex. V2) comes out, all of those update comments can be removed.

PR #24, opened 2020-05-15
ODT opt in json contract WEB-861

tkainz:
Initial Contract


odt-opt-in/initial ODT Opt-in JSON Contract.md
```diff
@@ -0,0 +1,155 @@
...
+
+Poweron response - With eligible shares:
+```json
+{
+ "memoMode": false,
+ "results": {
+  "maxSharesExceeded": false, ['true' if the number of shares found exceeds processing capbilities (130 shares)]
+  "shareDetail": [{           [note: the first share returned will be the share the member]
+    "SID": "0000",            [currently has opened, provided it is eligible. All other shares will be in heirarchal order]
+    "name": "My Primary Share"
```
jrobinson01, line 62: missing a comma at the end of this line and the next. Same for the second example share

odt-opt-in/initial ODT Opt-in JSON Contract.md
```diff
@@ -0,0 +1,155 @@
...
+    "currentState": true
+   }
+  ]
+ }
+}
+```
+
+Error Codes (individual error codes are for ease of researching issues.
+             All error codes will be returned to the UX as '500')
+500 - MemoMode Error
```
jrobinson01, line 151: I don't think we need the memo mode error specifically, since we already handle it with the memoMode boolean.

jrobinson01:
attempting to link this to the jira ticket: 
#closes WEB-861

odt-opt-in/initial ODT Opt-in JSON Contract.md
tkainz: My thoughts were as follows....
Even though we report the memo mode error via JSON through the use of a boolean, because it is still an error, for the purposes of internal code and consistencies, it was tied to a specific error code. If you'd still like me to change it, just let me know .

odt-opt-in/initial ODT Opt-in JSON Contract.md
```diff
@@ -0,0 +1,155 @@
...
+NOTES: every state checks memoMode - if memoMode is true, we cannot proceed.
+
+## PRELOADDATA
+To get things started, the client will send the initial PRELOADDATA request.
+```json
+{
+  "rgState": "PRELOADDATA",
+  "powerOnFileName": "BANNO.ODTOPTIN.V1.POW",
+  "userChrList": [
+    {"id": 1, "value": "1234567890S0010"},  [NOTE: 10-digit account+'S'+SHARE:ID]
```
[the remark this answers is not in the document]
tkainz: as follows?
{"id": 1, "value": "S0010"}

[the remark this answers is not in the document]
jrobinson01: ok let's leave it as is then. I probably won't be providing it in the first iteration of the ux but will look to adding it later.

odt-opt-in/initial ODT Opt-in JSON Contract.md
tkainz: corrected - waiting for clarification on other issues before posting update

odt-opt-in/initial ODT Opt-in JSON Contract.md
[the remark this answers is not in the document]
tkainz: As the current program logic is looking for it, if it's not there, there will be issues with the program.   If you don't currently have it in, then let's just leave it out altogether. I'll modify the code accordingly.

odt-opt-in/initial ODT Opt-in JSON Contract.md
```diff
@@ -148,7 +142,6 @@ PowerOn Response if successful
 
 Error Codes (individual error codes are for ease of researching issues.
              All error codes will be returned to the UX as '500')
```
jrobinson01, line 144: I think the 503 for "No eligible shares" should be passed through to the client. We have a specific error screen for that.

tkainz: Pushed new commit with updated error code reporting and to note that all error codes will be reported to UX as "500" with the exception of error code "503"

PR #36, opened 2020-08-11
Creating card block mgmt folder

tkainz:
Creating card block mgmt folder and adding initial JSON contract - (Work in progress)

### DIFF
--- a/card-block-mgmt/card-block-mgmt-json-contract.md
+++ b/card-block-mgmt/card-block-mgmt-json-contract.md
@@ -1,0 +1,147 @@
+# BANNO.CARD.LOC.MGMT.V1.POW JSON contract
+
+#### **Memo Mode Processing**
+As the PowerOn program principly modifies account tracking files, Should
+the system be in memo mode, program processing will end and error code
+500 will be returned. This applies to all states
+
+```json
+{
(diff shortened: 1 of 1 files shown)


card-block-mgmt/card-block-mgmt-json-contract.md
```diff
@@ -0,0 +1,167 @@
...
+{
+  "errorCode": 503,
+  "loggingErrorMessage": "Error processing request:"+[error]
+}
+```
+
+### Error codes
+| Code   | Description                        |
+|--------|------------------------------------|
+| 500    | System in memo mode                |
```
svetkomir, line 163: Why not start the error codes from `1`? To avoid confusion with the HTTP status codes.

tkainz: No good reason not to from my perspective. When I came aboard we were using 500's so I just kept going with that. Should we take it one step further and establish a series of error codes which we can consistently move forward with so an error code 1.. or 10.. or 100 for that matter would mean the same thing from one Banno program to the net?

svetkomir: Yep I like that.

PR #47, opened 2020-10-15
Update Certificate Renewal JSON contract.md

tkainz:
Added target account/share ID element during PROCESSDATA state. Seperated transferable shares from combined into separate groups, transferIn, transferOut.

### DIFF
--- a/certificate-renewal/Certificate Renewal JSON contract.md
+++ b/certificate-renewal/Certificate Renewal JSON contract.md
@@ -13,7 +13,7 @@
 }
 ```
  - userCharList[1]: 10-digit account number, 'S' and 4-digit share ID
-	 - Represents the target certificate or club share
+	 - Represents the certificate or club share being updated by the member
 
 
 ## Successful Response:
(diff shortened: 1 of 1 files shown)


certificate-renewal/Certificate Renewal JSON contract.md
```diff
@@ -162,8 +187,10 @@
 }
 ```
  - userCharList[1]: 10-digit account number, 'S' or 'L' and 4-digit share/Loan ID
+	 - Represents the certificate or club share being updated by the member
+ - userCharList[2]: 10-digit account number, 'S' or 'L' and 4-digit share/Loan ID
 	 - Represents the source or target share/loan of the transfer
- - userCharList[2]: User selected option
+ - userCharList[3]: User selected option
 	 - The transfer in/out amount
```
reuelcj, line 194: @tkainz the amount being transferred out is never calculated in the UI

tkainz: true....although I suppose the member 'could' do a partial transfer out of funds and then let the cert roll over... but that's not an option we're offering.

certificate-renewal/Certificate Renewal JSON contract.md
```diff
@@ -133,7 +152,12 @@
 	 - 5-**Disburse funds by check**: Withdraw balance by check, close certificate.
 	 - 6-**Suspend**: Suspend certificate activity - funds retained in share.
 
- - ***transferInOut***: A list of shares eligible for transferring certificate balance into or out of shares and loans as a source of funds for increasing certificate balance 
+ - ***transferIn***: A list of shares eligible as a source for transferring funds into the certificate upon maturity (transfer occurs the day before maturity.)
```
jrobinson01, line 155: minor nit - to stick with Banno convention/vocabulary, could these be changed to `transferTo` and `transferFrom`?

PR #49, opened 2020-10-16
Update BANNO.NEWSUBCREATE.V1.POW

christinasimms:
Did not merge last request as an issue was discovered. Tom fixed the issue and it has been retested.


open-subshare/REPWRITERSPECS/BANNO.NEWSUBCREATE.V1.POW
```diff
@@ -1028,7 +1073,7 @@ PROCEDURE SHAREOUTPUT
          IF J>0 THEN
           DO
            IF SACSHARETYPES(I,J-1)<>"" OR
-              SACSHARETYPEEXISTS(J-1)=TRUE THEN
+              SACSHARETYPEEXISTS(J-1)<>TRUE THEN
```
tkainz, line 1076: If I'm interpreting the code correctly, The SACSHARETYPEEXISTS flag is set when a test is done (lines 997-1001) to make sure the share types listed are defined.  These two lines of code are essentially looking to see that the previous one (J-1) is not blank and exists (is valid). Therefore, I'm not sure why the test to make sure that it would not be equal to true.... meaning that it doesn't exist. Is this being changed to "<>TRUE" here because, in effect, by removing - or in this case reversing - the test, we're correcting for an error in the code somewhere else? Would we get the same results by just using the "SACSHARETYPES(I,J-1)<>""" portion of the test?

open-subshare/REPWRITERSPECS/BANNO.NEWSUBCREATE.V1.POW
```diff
@@ -2164,4 +2238,31 @@ PROCEDURE JSONCLOSE
  NEWLINE
 END [PROCEDURE]
 
+PROCEDURE CUERRORHANDLING
+
+ [CHECK ERROR TEXT LENGTH]
```
tkainz, line 2243: Depending upon what the error being displayed is, could cutting it off at character 40 - no matter what - make it challenging to read?  Maybe consider starting at character 40 and backtracking until a space is found and then cutting the sentence at that point.

christinasimms:
Tom took over programming.

PR #50, opened 2020-11-04
[DOC DAY 2] Error Codes List

reuelcj:
Things to figure out ... 
What numbering system do we want to use?
What are pros/cons of ranges?
What groups would we use if we did ranges?

MDN on error handling: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status


reuelcj:
moved this to https://github.com/Banno/team-ux-admin/pull/81 as it doesn't really belong in this public repo and is more about how the UI works with the poweron

PR #63, opened 2021-02-11
ODT Optin terms JSON update

tkainz:
Updated JSON for terms, disclosure and revoke output to allow for blank lines - incl. first line. Added program revision info to JSON to assist with future issue research.


jucarson (approved):
Looks Good Tom

jordan-keenan (commented):
Part to Add program info into JSON Looks Good!
So it looks like you just removed the PRINT " " that is put after ',' between Lines that are not blank. 
Does the mean the JSON Contract no longer expects a " " after each comma before the textline so that it will count no " " after a comma followed by another comma as a blank line?

tkainz:
> Part to Add program info into JSON Looks Good!
> So it looks like you just removed the PRINT " " that is put after ',' between Lines that are not blank.
> Does the mean the JSON Contract no longer expects a " " after each comma before the textline so that it will count no " " after a comma followed by another comma as a blank line?



> Part to Add program info into JSON Looks Good!
> So it looks like you just removed the PRINT " " that is put after ',' between Lines that are not blank.
> Does the mean the JSON Contract no longer expects a " " after each comma before the textline so that it will count no " " after a comma followed by another comma as a blank line?

Jordan, in reviewing the original code, I cannot recall the why and where for of the original programming logic. There was  (must have been) a reason for why it was coded that way.  I can tell you that as it is - post changes - the program is no longer erring out when the first line is a blank line. a blank line in the config Letter file results in a hard return in the UX display and two blank lines back to back results in two hard returns (a paragraph).

jordan-keenan (approved):
Looks Good Tom!
I now see the change in the code that allows for blank lines in PARAMCUSTOMTERMS by printing them followed by a comma.
Whereas before it just printed nothing, and only added commas after non-blank lines.

PR #73, opened 2021-03-26
update m2m contract, again

jrobinson01:
Updates the m2m history events:
remove "MemberToMemberTransferSent" and replace with "MemberToMemberTransferScheduled" for the creation event
change `application` value to "member-to-member-transfers-poweron"


tkainz (commented):
Remember... the commit array index stops at 25 ;-)

PR #81, opened 2021-05-14
Update BANNO.LOAN.PAYOFF.V1.POW

tkainz:
1. Updated program header
2. Added check for duplicate loan types in different categories in CFG Letter file
3. Corrected issue where JSON output was including TOC detail when CFG File read error existed resulting in incorrect JSON
4. Added JSON output of CFG read info with 'timer' to shut off after 90 days
5. Corrected veh. collateral JSON info when corresponding tracking record field left blank


tkainz:
Peer review was completed by Justin Carson and approved via email.  Merging

PR #84, opened 2021-05-28
Database Check Scenario 0 update

tkainz:
Updated to add additional logic scenario and added debug code and program info JSON output


christinasimms (changes requested):
scenario 0 is creating a new tracking and it should be revising the existing

tkainz:
> scenario 0 is creating a new tracking and it should be revising the existing

Determined that Banno is not passing back the member's mbr num link and the addr num link as was expected. Awaiting review of new logic scenario 0 to determine what further changes are necessary.

tkainz:
@christinasimms With the current logic, 'Scenario 0' is never getting run as it requires the member link and address link values which are not (yet) being passed back in this state.

PR #88, opened 2021-07-14
Update BANNO.M2MTRANSFERS.V1.POW

tkainz:
Changed amount value in JSON History output to number from character (removed quotes - line 765). Also started cleaning up code, formatting.


jordan-keenan (approved):
Looks good! I see in the JSon contract it is looking for a number without quotation marks on each side, rather than a character field with quotation marks on each side. 
From JSon contract
 "amount": 1.00,

The code clean up also looks good!

PR #90, opened 2021-07-26
Update M2M Credit share requirements

tkainz:
Added additional requirements to validate credit shares (adding IRS Code and Share Code requirements)


jordan-keenan (approved):
The code looks good Tom.
I think what you are doing is making sure
a Savings will only work if it has a Share Code of 0.
And that you are also making sure the
CHECKING has and IRS CODE of 0 like you did
for SAVINGS. (I don't see how Certificate or Club would work, but maybe that is the point of the change).

jucarson (approved):
Looks Good!

PR #102, opened 2021-10-05
open-subshare add optional apy field

jrobinson01:


tkainz (commented):
Wouldn't the APY need to be at the interest rate level in the JSON code.  The rate would be different for each balance cutoff point and thus wouldn't the corresponding APY?

            "interestRates": [
              {
                "rate": "0.000",
                "APY": "0.000",
                "minBalance": "0.00",
                "maxBalance": "4.99"
              },

PR #105, opened 2021-11-02
Update BANNO.M2MTRANSFERS.V1.POW

tkainz:
Modified output format of the transfer record locator to eliminate any commas


jucarson (approved):
Looks Good.

jordan-keenan (approved):
Looks good.  Only thing I could see about transfer locator is it uses number which goes up to
2,147,483,647  (9 digits), which is one digit less than (8) than you have. But I don't think it would ever go that high as 9 digits - I could be wrong too about what you transfer locator refers too.

christinasimms (approved):

PR #107, opened 2021-11-16
Update to Banno ODT OP-in PowerOn CFG

tkainz:
Updated the Custom messages section to indicate the acceptable use of some HTML tags.


jordan-keenan (approved):
Looks good! I like the examples of the tags that would also work in a disclosure.
I suggest changing "FD:</ P>" to a lower-case p is the opening tag is lower-case. (Note: For this review comment, I had to add space after / before P for it to show up)

jucarson (approved):
Looks good!

PR #108, opened 2021-11-22
Created M2M Transfers Questionnaire

tkainz:
Created and added the M2M Transfers Questionnaire


jordan-keenan:
Does this actually need a code review? Let me know if I need to approve

tkainz:
@jordan-keenan A second set of eyes is always good whether looking for misspellings or just clarity of wording. The hope is to make sure all the necessary questions are asked while keeping them phrased properly and in such a way that, as much as possible, follow-up questions and clarifications can be kept to a minimum.

jordan-keenan (approved):
Looks Good. I suggestion that you can add if you want: for "Transaction Description/Comment Default" you can add an example of the most common phrase a CU would choose.

PR #111, opened 2021-12-08
Updates to M2M Transfers

tkainz:
Corrected transfer locator format issue in JSON output when locator is over 999.
Added S/L ID Length to JSON output.
Added procedure to normalize S/L ID length.
Added check for same account transfer (not allowed).
If tran counts are being enforced, and transfer initiated on system date is  deleted (expired) or edited.
Adjust transfer count and transfer amount total accordingly.
Added immed. transfer comment parameter.


jucarson (approved):
Looks good Tom!

jordan-keenan (approved):
The code looks very good. I just have a couple suggestions:

On line 1283 for debug, that uses NEWTRANSRCAMOUNT instead of describing it as NEWXFERCOUNT, I think you mean something like NEWTRANSRCAMOUNT?

IF “IL” is meant for CU’s to uses I suggest you adding “IL” (ID Length) to the CFG;
But if that is just for testing purposes, which it looks like it is, then no need to add it to the CFG.

tkainz:
> @jordan-keenan - line 1283 corrected - TY. "IL" is a 'hidden' parameter setting which was (can be) used to test the program with various Share/Loan ID length settings. It is not intended to be used by the CU.

PR #112, opened 2022-01-13
Added M2M Program overview

tkainz:
Added program overview to describe program functionality


jordan-keenan (approved):
This document is very professional. Good job Tom!

Thank you that you added the following at the bottom of the overview
document, as it is helpful:

"Individual member limitations will always override the global settings."

PR #120, opened 2022-03-17
Update BANNO.DATABASE.CHECK.V1.POW

tkainz:
Added additional scenario check (scenario 0)
Added program info and debug info to JSON output.
Removed procedures not being needed or called
Added additional documentation
Added scenario logic used to JSON output

### DIFF
--- a/link-user-to-episys-name-record/REPWRITERSPECS/BANNO.DATABASE.CHECK.V1.POW
+++ b/link-user-to-episys-name-record/REPWRITERSPECS/BANNO.DATABASE.CHECK.V1.POW
@@ -10,7 +10,8 @@
 **
 **  Modification History: (be sure to update the last mod date in the SETUP div. as well)
 **    Ver. 1.0.0  11/16/20 T. Kainz - Original Programming
-**    Ver. 1.0.1  03/17/22 T. Kainz - Added additional scenario check (scenario 0)
+**    Ver. 1.0.1  04/04/22 T. Kainz - Corrected Scenario 3 tracking record check & links check.
+**                                    Added additional scenario check (scenario 0)
 **                                    Added program info and debug info to JSON output.
 **                                    Removed procedures not being NEEDED OR CALLED
 **                                    Added additional documentation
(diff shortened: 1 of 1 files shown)


jucarson (approved):
Changes look great, Tom! Nice job on documentation!

link-user-to-episys-name-record/REPWRITERSPECS/BANNO.DATABASE.CHECK.V1.POW
```diff
@@ -969,7 +1134,15 @@ PROCEDURE POLLTRACKING
...
 
+[* USERCHAR12 is the preference Locator and connects a specific tracking 8 to
+** the member logged in. USERCHAR12 is only filled in for BANNO use.
+*]
+
+ BNODEBUGMSG="ACCOUNT: "+ACCOUNT:NUMBER
+ CALL BNODEBUGMSGADD
+
  FOR EACH TRACKING WITH (TRACKING:TYPE=8 AND
+                         VALUE(TRACKING:USERCHAR12)>0 AND
```
jordan-keenan, line 1155: Hi Tom,
 I think adding the check for VALUE(TRACKING:USERCHAR12)>0 before loading the tracking in the tracking array to be evaluated will prevent NetTeller Tracking records from getting updated with banno log in information. I remember the Idea was to recycle NetTeller tracking 8 records. NetTeller stores that Value for Preference Locator in 
USERNUMBER1

jordan-keenan: Looking at it more deeply, it appears the old code was looking for a matching TRACKING:USERCHAR12, and if not found it just created a tracking 8. So it looks like we have been making Banno tracking 8, regardless of  if they have a NetTeller tracking 8. Is that better in some way?

link-user-to-episys-name-record/REPWRITERSPECS/BANNO.DATABASE.CHECK.V1.POW
```diff
@@ -765,7 +951,11 @@ PROCEDURE GETNAMETYPES
 ** second in order of name type.
 *]
  VALIDNAMETYPECOUNT=0
- LELISTINPUT=@RGUSERCHR1
+ TMPCHR=@RGUSERCHR1
+ CALL NLTS
+ LELISTINPUT=TMPCHR
+ IF LELISTINPUT="" THEN
+  LELISTINPUT="ALL"
```
jordan-keenan, line 958: Not sure if we want ALL name types if they send us nothing. In the Banno DB though if it is blank right now they are sending us a name type list of 0,1,5-9 per what they said the default is for name types in documentation. Not sure if this should be change to "0,1,5-9" if it comes as blank, or if Leaving it as ALL is fine since as of now they will send us a specific list, or they will send us 0,1,5-9 if the Banno DB has no values.

link-user-to-episys-name-record/REPWRITERSPECS/BANNO.DATABASE.CHECK.V1.POW
```diff
@@ -480,6 +642,7 @@ SUCCESS
            IF MBRLINKFOUND=TRUE OR
```
jordan-keenan, line 642: Overall the improvements to the code are very helpful to have Episys Notes and Debug Code to see in the Logs! Thank you for all the work Tom. 

I think I found a minor issue in the logic of part of Scenario 3.
I think you mean "IF MBRLINKFOUND=TRUE **AND** ADDRLINKFOUND=TRUE THEN"  Instead of OR (meaning one name record has match one and another name record has match for the other) on line 642 for LINKTRACKING Logic Scenario #3.c.2, as The conditions statements above where it "says 3.c.1. If both numbers are found in a single name record, return success." 

Otherwise I think none of the code below #3.c.2 would happen, which is designed to handle the situation of having found only as single name that either contains the matching MBRLINK or the matching ADDRLINK.

I also, suggest promoting scenario 3 to the top of LINKSTATE, as 3.c.1 not needing to update tracking 8, would eventually be the most common scenario, unless the more common scenario is Members with both a Prime and Joint Name, making the error for multiple names .

Please take look at the other comments/suggestions I left that are not part of the code review.
Let me know if I am mistaken somehow about the rest of the Scenarios in 3 after #3.c.2 not being able to happen.

Thank you,

jordan-keenan (changes requested):
Overall the improvements to the code are very helpful to have Episys Notes and Debug Code to see in the Logs! Thank you for all the work Tom.

I think I found a minor issue in the logic of part of Scenario 3.
I think you mean "IF MBRLINKFOUND=TRUE AND ADDRLINKFOUND=TRUE THEN" Instead of OR (meaning one name record has match one and another name record has match for the other) on line 642 for LINKTRACKING Logic Scenario https://github.com/Banno/banno-powerons/pull/3.c.2, as The conditions statements above where it "says 3.c.1. If both numbers are found in a single name record, return success."

Otherwise I think none of the code below https://github.com/Banno/banno-powerons/pull/3.c.2 would happen, which is designed to handle the situation of having found only as single name that either contains the matching MBRLINK or the matching ADDRLINK.

I also, suggest promoting scenario 3 to the top of LINKSTATE, as 3.c.1 not needing to update tracking 8, would eventually be the most common scenario, unless the more common scenario is Members with both a Prime and Joint Name, making the error for multiple names .

Please take look at the other comments/suggestions I left that are not part of the code review.
Let me know if I am mistaken somehow about the rest of the Scenarios in 3 after https://github.com/Banno/banno-powerons/pull/3.c.2 not being able to happen.

Thank you,

link-user-to-episys-name-record/REPWRITERSPECS/BANNO.DATABASE.CHECK.V1.POW
tkainz: This was really added in as a stop-gap.  We found that there were times when Banno was sending back a null string in which case the listexpand code was interpreting that as a "none". Better to allow ALL name records instead of none.

link-user-to-episys-name-record/REPWRITERSPECS/BANNO.DATABASE.CHECK.V1.POW
tkainz: Jordan.... I'm not sure about your comment on 3.c.1 & 3.c.2.

3.c.1 is looking for a single name record where both the member link and the address link match. IF there is no name record where both the name and address links match then the next check is 3.c.2:  the program then looks to see if there is at least one record where either the address link OR the name link matches.  Call me if you'd like to discuss.

jordan-keenan (approved):

jordan-keenan:
The code looks good Tom! I approved the changes. Looks like you fixed more bugs that I didn't see.

I noticed these two lines are removed from the previous commit:
 PROGRAMUPDATENOTE1="Added Program info and @RG detail to JSON Output. Added Scenario 0."
 PROGRAMUPDATENOTE2="Added additional error handling when Tracking 8 is not being created"

Is that because they were not sent in the print out? That is fine to leave them out; However; I do see how if they where left in the code and printed out, they would be helpful when looking at the logs.

Thanks,

tkainz:
> The code looks good Tom! I approved the changes. Looks like you fixed more bugs that I didn't see.
> 
> I noticed these two lines are removed from the previous commit: PROGRAMUPDATENOTE1="Added Program info and @rg detail to JSON Output. Added Scenario 0." PROGRAMUPDATENOTE2="Added additional error handling when Tracking 8 is not being created"
> 
> Is that because they were not sent in the print out? That is fine to leave them out; However; I do see how if they where left in the code and printed out, they would be helpful when looking at the logs.
> 
> Thanks,

@jordan-keenan I deleted those two lines because they were never used by the program (as part of the JSON Output).  I also figured that as long as the program version number and last mod date was part of the JSON output, then we'd be able to look at the last modification date in the program header to see what those changes included.

jucarson (approved):

PR #123, opened 2022-04-06
BANNO.NEWSUBCREATE.V1.CONFIG HTML Update

tkainz:
Updated to add additional information to the template Terms Letter files when they are created by the program. The new Template terms Letter files created by the program (as needed) will appear as follows
```
Sample Terms for Banno Sub Acct Opening


Data will be displayed as written. The following HTML tags may be utilized to facilitate text formatting:
    h1, h2, h3, h4, h5, h6, b, strong, i, u, ul, ol, li, br, p, a and hr.
Of course... don't forget to include them in their required delimiters '<>'. Avoid the use of double quotes
or special characters in your message.


For backwards compatibility, a single blank line will be interpreted as a new line and two blank lines in a row
will be interpreted as a new paragraph


Use this ruler as a guideline when setting up your messages. Do not extend past the end of this line (120 characters).


         10        20        30        40        50        60        70        80        90        100       110       120
123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890
---------|---------|---------|---------|---------|---------|---------|---------|---------|---------|---------|---------|

<h3>Savings Shares</h3>
<ul>
<li>Condition 1</li>
<li>Condition 2</li>
</ul>
```

### DIFF
--- a/open-subshare/REPWRITERSPECS/BANNO.NEWSUBCREATE.V1.CONFIG
+++ b/open-subshare/REPWRITERSPECS/BANNO.NEWSUBCREATE.V1.CONFIG
@@ -3,12 +3,15 @@
 **  Letterfile Name:    BANNO.NEWSUBCREATE.V1.CFG
 **                      BANNO.NEWSUBCREATE.TERMS.00-09
 **
-**  Copyright 2020 Jack Henry and Associates
+**  Copyright 2020-2022 Jack Henry and Associates
 **
 **  05/18/2020: Initial program creation
 **  12/17/2020: TKainz: Added check for share having a defined
 **              dividend type. If undefined, share type is not
 **              eligible for selection.
+**  04/05/2022: TKainz: Updated sample text and instructions in
+**              BANNO.NEWSUBCREATE.TERMS.xx Letter files to include
+**              the ability to use basic HTML commands.
 **
 **  This Banno service PowerOn is to be installed for use on
 **  demand and is run from the Account Manager workspace to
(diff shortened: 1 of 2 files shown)


open-subshare/REPWRITERSPECS/BANNO.NEWSUBCREATE.V1.CONFIG
```diff
@@ -1140,14 +1143,44 @@ PROCEDURE CREATETOCLETTERFILES
...
+*]
      FILEOPEN("LETTER",TMPCHR,"WRITE",FILENUMBER,FILEERROR)
      IF FILEERROR="" THEN
       DO
        FILEWRITELINE(FILENUMBER,"Sample Terms for Banno Sub Acct Opening",FILEERROR)
        FILEWRITELINE(FILENUMBER,"",FILEERROR)
-       FILEWRITELINE(FILENUMBER,"Certificates Shares...",FILEERROR)
-       FILEWRITELINE(FILENUMBER,"Condition 1:",FILEERROR)
-       FILEWRITELINE(FILENUMBER,"Condition 2:",FILEERROR)
+       FILEWRITELINE(FILENUMBER,"",FILEERROR)
```
jordan-keenan, line 1153: Hi Tom, This looks exactly what we want! 
However; can you make some way to have these comments ignored by the UX, so the CU's can leave them there? That way when they change terms later, the comments about html tags will still be there.

As it is now we, meaning Justin or I, some future Person who installs Poweron, or the CU will delete the comments immediately to get something going, and if they didn't put in HTML and want it later those comments explaining it will be missing.

tkainz: Jordan,
I'd have to add code to both programs (the NEWSUBCREATE.POW & NEWSUBCREATE.CONFIG) to use the same logic which is used in most of the other programs where any lines starting with a certain character combination will be ignored by the NEWSUBCREATE.POW program when being read in and immediately written out to the JSON.  We'll be taking a risk in that any TOC file already out there which may already have a TOC file which has a line or more with that combination will have their output ignored by the program.  What character combination do you recommend which the program could use to ignore a TOC Letter file line which is most likely not possibly already used in an existing Letter file? I would fear that if we were to use the standard asterisk ('*') as the first character in the line, that any number of CU's may have already used that character in their standard TOC to represent a bullet point type item.

jucarson: I agree with both points. I feel that the amount of work required to accommodate that isn't worth the benefit. I'm approving for now, if you guys decide collectively to change them, I will rereview.

jucarson (approved):

open-subshare/REPWRITERSPECS/BANNO.NEWSUBCREATE.V1.CONFIG
jordan-keenan: Hi Tom,
 I like that you have a way to make the change.
I’m thinking FI’s that already have BANNO.NEWSUBCREATE.V1.CONFIG are unlikely to update to the new one for a while; so I’m thinking there may not be too many of those situations as it would only be if the first character of a line is * .
And we could keep in mind to check for that if any support cases come in about the Subaccount create Letterfile not displaying a line. It may make sense to output the first line that begins with a * in the debug in the first iteration of that code, so if a CU upgrades, we could see the issue in traces (but would probably want to remove that debug line down the road).

I guess ‘[‘ is an alternative but could be more confusing as they might think it terminates with ‘]’.

 I’m also OK, if you decide not to do it. I’m just thinking now is the time to make the change while we are adding comments about HTML to the Letterfiles.

Let me know what you decide.

tkainz: I can see the reason for wanting to have the ability to include "instructions" in the header of the sample Letter files so the CU could refer to them as they update the respective Letter files.  That being said, These "instructions" would only benefit CUs whom are setting this up for the first time as the program will not overwrite existing Letter files.  For those CU's whom are setting this up for the first time, the instructions are being printed in the top of each of the template Letter files that the program is creating thus giving the information the CUs need to take advantage of the new HTML capabilities.

So, for now, I think I'll leave the changes to the template Letter files as they currently are and not add the ability to leave comments (instructions) in each of the Letter files and instead, update the README.md file to include the instructions. The CUs, after all, should be referring to that file when implementing the program use.

jordan-keenan:
Thanks for adding the comments about the HTML tags to the readme.

jordan-keenan (approved):
Thanks for adding the comments about the HTML tags to the readme.

PR #129, opened 2022-06-16
Update m2m-transfers-json-contract.md

tkainz:
Added Share flag to indicate if share is a cross-account. Clarified definition and UX submission requirement for an immediate transfer request.


tkainz:
Duplicate

PR #134, opened 2022-08-04
Loan Payoff fee display feature

rrobison44:
New code for [CUS-1766](https://banno-jha.atlassian.net/browse/CUS-1766)
Via a subroutine specfile, it is now possible to return up to 10 additional fees that will display in the UX.

- Updated Loan Payoff PowerOn JSON contract
- Updated library version of BANNO.LOAN.PAYOFF.V1.CFG configuration letterfile
- Updated library version of BANNO.LOAN.PAYOFF.V1.POW PowerOn code

Changes include:
- New option to specify a subroutine specfile for the purpose of returning fees to the UX for display.

#### Example screenshot
image


loan-payoff/REPWRITERSPECS/BANNO.LOAN.PAYOFF.V1.POW
```diff
@@ -833,6 +867,84 @@ PROCEDURE DISPLAYRESULTS
...
  CALL GETCOLLATERAL
+
+ IF PAYOFFFEECOUNT>0 THEN
+  DO
+   PRINT "   "+Q+"payoffFees"+Q+":["
+   NEWLINE
+   FOR TMPLOOP=1 TO PAYOFFFEECOUNT
+    DO
+     TMPCHR=FORMAT("#######9.99",PAYOFFFEEAMT(TMPLOOP))
+     CALL REMOVELEADINGSPACES
```
tkainz, line 878: (Realize that this is not relevant to your coding changes) - I noticed that there are two procedures for removing leading spaces (REMOVELEADINGSPACES & TMPCHRNLS) Can we choose one to use and get rid of the other?  Also, for the related procedure TMPCHRNTS... can that just be reduced to

```
PROCEDURE TMPCHRNTS
 TMPCHR=SEGMENT(TMPCHR,1,LENGTH(TMPCHR))
END
```

loan-payoff/REPWRITERSPECS/BANNO.LOAN.PAYOFF.FEES.POW
```diff
@@ -0,0 +1,195 @@
+[*
+**  PowerOn Name:       BANNO.LOAN.PAYOFF.FEES.POW
+**
+**  Copyright 2022 Jack Henry and Associates
+**
+**  This Banno subroutine PowerOn is called by the main BANNO.LOAN.PAYOFF.V1.POW PowerOn -- if configured
+**  in the CFG letterfile. It can be modified to return custom fee amounts and descriptions to the UX for display.
```
tkainz, line 7: Probably better to specify the entire PO filename here instead of just 'CFG letterfile' to avoid any ambiguities.

loan-payoff/REPWRITERSPECS/BANNO.LOAN.PAYOFF.FEES.POW
```diff
@@ -0,0 +1,195 @@
...
+
+ CALL ENVPARAMSINIT
+ CALL PAYOFFFEEINIT
+END
+
+PRINT TITLE="BANNO.LOAN.PAYOFF.FEES.POW"
+
+ FOR ACCOUNT PAYOFFACCOUNTNUM
+  DO
+[*
```
tkainz, line 79: Since this subroutine Specfile pertains to loans, it might be good here to add the FOR EACH LOAN WITH (LOAN:ID=PAYOFFSLID) clause to better illustrate what the program will actually be doing.

loan-payoff/REPWRITERSPECS/BANNO.LOAN.PAYOFF.V1.POW
```diff
@@ -31,6 +32,7 @@
 **                         Added fixes for invalid punctuation in disclosure 
 **                         Corrected processing of disclosure blank lines
 **                         Added forced space between consecutive lines of text 
+**    Ver. 1.0.4  08/03/22 R. Robison - Added support for payoff fee display in the UX.
```
tkainz, line 35: As this is an enhancement and not a bug fix, wouldn't the next version be 1.1.0?

loan-payoff/LETTERSPECS/BANNO.LOAN.PAYOFF.V1.CFG
```diff
@@ -93,6 +93,15 @@ Model:USERCHAR2
 *  assumed valid. ex. 0,10-20
 SLT:46
 *
+*
+* FEE SPECFILE NAME
+*  Line must begin with "FSN:" 
+*  Enter the name of the subroutine specfile that is responsible for returning loan payoff fees.
+*  If no fee specfile is entered, no additional fees will be displayed in the UX on the
+*  loan payoff screen.  Requirements are detailed in template specfile BANNO.LOAN.PAYOFF.FEE.POW.
```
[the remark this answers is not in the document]
rrobison44: Yip.  Fixed.

loan-payoff/REPWRITERSPECS/BANNO.LOAN.PAYOFF.V1.POW
```diff
@@ -172,9 +190,9 @@ SETUP
 
  VALIDCHRINPUT1="32-126,128-175,181-184,189,190,198,199,208-216,221,222,224-253"
 
- PROGRAMVERSION="1.0.3"
- LASTMODDATE='11/02/21'
- LASTMODTIME="16:00 PT"
+ PROGRAMVERSION="1.0.4"
```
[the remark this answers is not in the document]
rrobison44: I believe so.  Set version to 1.1.0

loan-payoff/REPWRITERSPECS/BANNO.LOAN.PAYOFF.V1.POW
```diff
@@ -1,8 +1,9 @@
 [*
-**  PowerOn Name:       BANNO.LOAN.PAYOFF.V1.POW
-**  Letterfile Name:    BANNO.LOAN.PAYOFF.V1.CFG
+**  PowerOn Name:             BANNO.LOAN.PAYOFF.V1.POW
+**  Letterfile Name:          BANNO.LOAN.PAYOFF.V1.CFG
+**  Fee Display PowerOn Name: BANNO.LOAN.PAYOFF.FEES.POW (template)
 **
-**  Copyright 2019 Jack Henry and Associates
+**  Copyright 2022 Jack Henry and Associates
```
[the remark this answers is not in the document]
rrobison44: Yes.  Fixed.

loan-payoff/loan-payoff-json-contract.md
```diff
@@ -1,49 +1,57 @@
 # Loan payoff JSON contract
+
 To get things started, the client will send an initial request structured like so:
+
 ```json
 {
   "rgState": "GETPRELOADDATA",
   "powerOnFilename": "BANNO.LOANPAYOFF.POWERON.V0.POW",
```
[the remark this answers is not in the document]
rrobison44: I did notice this...should have fixed it :). Fixed now.

loan-payoff/REPWRITERSPECS/BANNO.LOAN.PAYOFF.V1.POW
```diff
@@ -595,6 +612,15 @@ PROCEDURE READCONFIGFILESETTINGS
           END
         END
 
+       [ Fee specfile name]
+       IF SEGMENT(LFLINE,1,4)="FSN:" THEN
+        DO
+         TMPCHR=SEGMENT(LFLINE,5,LENGTH(LFLINE))
+         CALL TMPCHRNLTS
+         FSNCHR=TMPCHR
```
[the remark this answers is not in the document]
rrobison44: I was just trying to be consistent with the model that was already used.  Seems like there are a bunch of those xxxCHR variables that are just used to display the config info in procedure DISPLAYCONFIGINFO.  So I think I will leave it this was just for consistency. Code-wise, I agree it is not great.  A bunch of that code could probably be reworked.

loan-payoff/REPWRITERSPECS/BANNO.LOAN.PAYOFF.V1.POW
```diff
@@ -833,6 +867,84 @@ PROCEDURE DISPLAYRESULTS
...
+ IF PAYOFFFEECOUNT>0 THEN
+  DO
+   PRINT "   "+Q+"payoffFees"+Q+":["
+   NEWLINE
+   FOR TMPLOOP=1 TO PAYOFFFEECOUNT
+    DO
+     TMPCHR=FORMAT("#######9.99",PAYOFFFEEAMT(TMPLOOP))
+     CALL REMOVELEADINGSPACES
+     PRINT "    {"+Q+"feeAmount"+Q+":"+Q+TMPCHR+Q+","
+     PRINT Q+"feeDescription"+Q+":"+Q+PAYOFFFEEDESC(TMPLOOP)+Q+"},"
```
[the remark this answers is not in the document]
rrobison44: I think so.  I was a little confused because in code, this is ok but I think the JSON will not like it.  I can't even save it with a trailing space to mock test it :).  Fixed.

loan-payoff/REPWRITERSPECS/BANNO.LOAN.PAYOFF.V1.POW
rrobison44: I'm not sure about the TMPCHRNTS question.  Does the LENGTH function not consider trailing spaces when determining the length?  Your proposed change does simplify things quite a bit if the above is true.

On REMOVELEADINGSPACES and TMPCHRNLS, it would be nice to eliminate one.  Those procedures, logically, do approach the removing of leading spaces a little different.  Do we feel comfortable that they produce the same results?  I don't recall the specifics but seems like there was something weird/tricky about how leading/trailing spaces and the LENGTH function.  So I guess I am on the fence with regards to changing this code.  What are your thoughts?

loan-payoff/REPWRITERSPECS/BANNO.LOAN.PAYOFF.FEES.POW
rrobison44: Done.

loan-payoff/REPWRITERSPECS/BANNO.LOAN.PAYOFF.FEES.POW
rrobison44: Good idea.  Changes made.

loan-payoff/REPWRITERSPECS/BANNO.LOAN.PAYOFF.V1.POW
rrobison44: Yes.  Made this change.

loan-payoff/REPWRITERSPECS/BANNO.LOAN.PAYOFF.V1.POW
tkainz: In walking through the two variations of the code to remove leading spaces, it seems that they're both ending up with the same results - albeit it, like you noted, in slightly different ways.  The LENGTH statement does not take into consideration any trailing spaces (according to eDocs - `Counts leading, not trailing, spaces in the length`). Thus "`Tom   '" (which has three spaces at the end) should be returning a length of 3. Also, because both versions of the procedures to remove leading spaces reference the LENGTH function, they also, inadvertently, removing trailing spaces as well. 

In looking at the code, however for the two TMPCHR... procedures, I'm thinking that they both might fail IF the value they're testing is all spaces.  Given a TMPCHR value of `"   "` (three spaces) each of those procedures first start by testing for a LENGTH(TMPCHR) being greater than 0.  Since the LENGTH function does not count trailing spaces, in this case it "should" return a length of 0. Since it fails the LENGTH(TMPCHR)>0 test, all the code in the DO/END clause should be getting skipped over and thus, TMPCHR would get left with a value of 3 spaces... and thus, would also not be blank... technically.

tkainz (commented):
Added more thoughts on the leading / trailing spaces functions

loan-payoff/REPWRITERSPECS/BANNO.LOAN.PAYOFF.V1.POW
jucarson: I think it'd be best to clean it up a little and take one out given how eDocs says the LENGTH function works. But it also goes back to Robby's previous note that there's a lot that can be re-worked in the original code. As far as it failing on spaces, I think it'd be best to just double check where the procedures are called and ensure they'll have data besides a space. It's probably not an issue since they seem to be primarily used to remove spaces after a format, and all the formats have 9's, so at least a zero will be passed.

jucarson (changes requested):
Changes all look good. I just suggest changing the header in the 2 PowerOns and the CFG file from "Jack Henry & Associates" to "Jack Henry".

PR #138, opened 2022-09-20
cd renew terms

rrobison44:
The CD Renew PowerOn was not was not always displaying the correct renew terms for certain types of certificate/club shares.  To better support these types of accounts, the `currentTerms` preload variable is being replaced by:

```
renewName
renewTermPeriod
renewTermFrequency 
```

The UX code will continue to work for older version of the PowerOn code that only return `currentTerms`.

The `currentTerms` preload variable has been eliminated.  The UX code does not display any current terms but, instead, only displays renewal terms.  The renaming/new variables make more sense because they are used by the UX exclusively for certificate/club renewals.

[Jira Story CUS-1901](https://banno-jha.atlassian.net/browse/CUS-1901)
[Related UX PR](https://github.com/Banno/power-on-ux/pull/691)


tkainz (approved):
not seeing anything needing changes - Thanks

PR #139, opened 2022-09-22
Update Address Change

jucarson:
Update name matching logic.


jordan-keenan (approved):
Hi Justin, the Code Looks good, done well.

I suggest some comments (part in CAPS to call ATTENTION) added to the CFG Matching Parameter with this code review about what to do to the database check PO if they previously used what was in the tracking 8 records for Names linking in the Banno Address Change PowerOn, but are enabling matching instead. 

If you are going to make a PO that lets them “take care” of updating the tracking 8’s first, I suggest mentioning the name of that too.

Or if that will be part of this PO, I suggest mentioning what config settings to look at in that situation – not sure if that would be a separate review if you choose that option or if that would just be added to this review.

change-address/REPWRITERSPECS/BANNO.CHANGE.ADDR.V1.POW
```diff
@@ -709,14 +712,25 @@ PROCEDURE GETNAMELOCS
...
+  DO
+   OLDSTREET=NAME:STREET
+   OLDEXTRA=NAME:EXTRAADDRESS
+   OLDCITY=NAME:CITY
+   OLDSTATE=NAME:STATE
+   OLDZIP=NAME:ZIPCODE
+  END
  IF MATCHACCTNAMES=TRUE THEN
   DO
    FOR EACH NAME WITH ((NAME:EXPIRATIONDATE=DATENULL OR
```
tkainz, line 725: Should the check for name record qualifications include the NAME:EFFECTIVEDATE?  Maybe not.... just want to make sure it was considered.

jucarson:
> 

Thanks Jordan! I am going to merge this now for the CU's that are waiting for it. When I complete the FM PowerOn, I will open a new PR, add the comments in caps and work with Emily on being clear in the overview.

change-address/REPWRITERSPECS/BANNO.CHANGE.ADDR.V1.POW
```diff
@@ -709,14 +712,25 @@ PROCEDURE GETNAMELOCS
 ** information for allowed name types and levels. Only
 ** called if old address matching is enabled.
 *]
+ FOR EACH NAME WITH (NAME:TYPE=0)
+  DO
+   OLDSTREET=NAME:STREET
+   OLDEXTRA=NAME:EXTRAADDRESS
+   OLDCITY=NAME:CITY
+   OLDSTATE=NAME:STATE
+   OLDZIP=NAME:ZIPCODE
```
tkainz, line 721: Since everywhere (that I can see) is comparing only the first 5 characters of the 'OLDZIP', why not just store the first 5 characters here and eliminate the need for the system to recalculate the "SEGMENT(OLDZIP,1,5)" value multiple times below (for each found name record)?

PR #142, opened 2022-10-10
NEWSUBCREATE add name copy enhancement & other updates

tkainz:
Change overview
1.	Allow users to select existing names to be copied to the new share record being created. a.	Share and loan name types eligible to be copied over are controlled by the new configuration program settings i.	Copy share names flag & eligible name types
ii.	Copy loan names flag & loan name types
b.	While the program builds an internal list of name records which can be copied, based upon the config settings, it is also checking for duplicates (based upon the SSN field in the name records.) i.	If the program encounters multiple name records with the same SSN, the first one found will be presented to the member as an eligible option to copy.
2.	Add “testing” function to the program a.	Allow the CU to turn on test mode and set up to 10 test accounts. b.	With test mode on if a member is one of the test accounts, then they will be able to test some of the new functionality whereas other members will be limited to current functionality. c.	New functionality the test accounts will have access to are: i.	Ability to copy existing name records to the new share being created ii.	Ability to select a new share from up to 20 defined groups instead of 10.
3.	Add CU defined error code descriptions a.	These changes greatly reduce the possibility of the member getting an “oops” error. Instead, for most error conditions which can occur before the new share is created, the member would be presented with an error message better clarifying what the error was. These 11 error messages can be defined by the CU via the configuration program.
4.	Added ability for the CU to add comment lines to the Terms or Conditions Letter files.
5.	Corrected a few JSON issues
6.	Corrected issues with the configuration program’s form layout keeping it from properly and accurately being rendered.
7.	Added ability to force error conditions (1 through 11) to facilitate testing program error conditions a.	Requires configuration program to be in debug mode and error code entered into the test error code input field
8.	Corrected how name creation error and share transfer errors were being handled. (they weren’t)


open-subshare/REPWRITERSPECS/BANNO.NEWSUBCREATE.V1.POW
```diff
@@ -826,23 +1188,19 @@ PROCEDURE QUALIFYMEMBER
...
     LELISTINPUT="none"
    CALL LISTEXPAND
-   IF LELIST(ACCOUNT:TYPE)=TRUE THEN
+   IF LELIST(ACCOUNT:TYPE)=TRUE OR
+      TESTERRORCODE=503 THEN
     DO
      PRINT "{"
      NEWLINE
      CALL PRINTPROGRAMINFO
-     PRINT "  "+Q+"canOpenSubShare"+Q+": false,"
```
jucarson, line 834: Hey Tom, When testing my changes to this PowerOn for Deere, Emily noticed that if the account was an invalid account type, she would get a "Cannot Open Subshare" message, but when she had a warning that matched an account level warning in the parameters, she would get an "Oops" Message. I found it was because this line I am writing the comment on is missing above in the warning check. I wanted to add a comment to it before I really PR'd it so you could add that change. BUT when I got to this point in the new code to write this comment. I notice you set the "canOpenSubShare" to true. Shouldn't it, and the new line I am suggesting, be set to false?

tkainz: @jucarson annnnnnddddd I have no clue at this point as to why it was changed from false to true - maybe a search/replace gone wild thing??  With the changes to this version of the code, however, the 'canOpenSubshare' value should be a moot point.   The error code is being set to a 503 value and the error handling procedure called where the displayErrorMessage is being set. It's the displayErrorMessage which is then being intercepted by the UX and the error message displayed to the end user and program run effectively halted instead of the 'canOpenSubshare: false' trigger. That being said, I did correct the issues you noted for error codes 502/503 to make them consistent.  This did, however, also allow me to catch a type-o related to the max error code (corrected '11' to '511'.  Thx!

open-subshare/REPWRITERSPECS/BANNO.NEWSUBCREATE.V1.CONFIG
```diff
@@ -1,17 +1,34 @@
 [*
 **  PowerOn Name:       BANNO.NEWSUBCREATE.V1.CONFIG
 **  Letterfile Name:    BANNO.NEWSUBCREATE.V1.CFG
```
jordan-keenan, line 3: I suggest changing the comment on line 3 "Letterfile Name:" to "Datafile Name:"

open-subshare/REPWRITERSPECS/BANNO.NEWSUBCREATE.V1.POW
```diff
@@ -2416,46 +3072,44 @@ PROCEDURE ERRORHANDLER
 **  509 - Error getting new share rate
 **  510 - Could not calculate new maturity date
 **  511 - Error creating new share
-**  512 - Error creating new name record(s)
```
jordan-keenan, line 2419: Did you mean to remove Error Code 512 and 513? And if so does
open-subshare-json-contract.md need to be changed? Or will it be left alone in case you use 512 or 513 later?

were removed in new version:
512 - Error creating new name record(s)
513 - Transfer funding error

open-subshare/REPWRITERSPECS/BANNO.NEWSUBCREATE.V1.CONFIG
tkainz: Makes sense - will do

open-subshare/REPWRITERSPECS/BANNO.NEWSUBCREATE.V1.POW
tkainz: While reviewing the code for these changes I discovered that error codes 512 & 513 were not being processed in the same manner as the other error codes. This is due to the fact that by the time a 512 or 513 occur, the share has already been created and thus, program halt cannot occur.  512 is being handled by passing the "created:[true/false]" being passed to the UX for each name creation attempt and code 513 is handled by passing the "successfulTransfer:[true/false]" value being passed to the UX.

Yes, the contract will need to be / should be updated.

open-subshare/REPWRITERSPECS/BANNO.NEWSUBCREATE.V1.POW
```diff
@@ -1,7 +1,7 @@
 [*
 **  PowerOn Name:       BANNO.NEWSUBCREATE.V1.POW
 **  Datafile Name:      BANNO.NEWSUBCREATE.V1.CFG
-**  Letterfile Name:    BANNO.NEWSUBCREATE.TERMS.00-09
+**  Letterfile Name:    BANNO.NEWSUBCREATE.TERMS.00-19
 **
 **  Copyright 2020-2022 Jack Henry and Associates
 **
```
jordan-keenan, line 7: I was a little thrown off at at line 2100, as there is the only remaining "CALL JSONCLOSE"
in the program. I think it might be correct to put it there, as you do another 
open bracket after that and then do PRINTPROGRAMINFO.

I had just noticed you removed all the other "CALL JSONCLOSE" parts of the code as 
you have added the following to the end of the
print division, PRINT BNOINDENT(0)+"}" .

tkainz: Yes, instead of having a JSONCLOSE at the end of each (many - actually) procedures, it was just added to the end of the print division. As for the JSONCLOSE calls, there are two remaining and thus the procedure call instead of imbedding the actual JSON closing brace into the code directly.

jucarson (commented):
Code looks good. Just had 2 minor suggestions/questions... 

On the worksheet.. I suggest having a separate copy share names types and copy loan names types, so they realize that there can be separate answers for each type. Also, in the "Share Groups" heading, can you change 10 to 20 groups of shares and add a note to copy and paste the tables for share groups 9-20?

open-subshare/REPWRITERSPECS/BANNO.NEWSUBCREATE.V1.POW
```diff
@@ -323,22 +330,219 @@ DEFINE
...
+ TESTNAMESTREET            = CHARACTER
+ TESTNAMEZIPCODE           = CHARACTER
+ TMPLOC                    = NUMBER
+ TMPNUM                    = NUMBER
+ TRUEFALSECHR              = CHARACTER ARRAY(1)
+ VALIDLOANNAMETYPES        = NUMBER ARRAY(99)
+ VALIDSHARENAMETYPES       = NUMBER ARRAY(99)
+ VALIDNAMETYPESCHR         = CHARACTER
+ VALIDNAMETYPES            = NUMBER ARRAY(99)
+
```
jordan-keenan, line 451: I don't see VALIDNAMETYPESCHR and VALIDNAMETYPES used in the program. Are they for future use?

Note: I see this VALIDNAMETYPESCHR="ALL"  in code, but I don't see it used.

open-subshare/REPWRITERSPECS/BANNO.NEWSUBCREATE.V1.POW
```diff
@@ -546,34 +764,77 @@ PRINT TITLE="BANNO.NEWSUBCREATE.V1.POW"
...
        PRINT "   "+Q+"canOpenSubShare"+Q+": true,"
        NEWLINE
+
+[* RESULTS START
+*]
        PRINT "  "+Q+"results"+Q+": {"
        NEWLINE
-
+[* if new names are not allowed or if they are allowed but the max new names
+** are nset to 0 or and there are no eligible copy names then set new names
```
jordan-keenan, line 773: I think you mean to say "are set to 0" instead of "nset" in the comment "but the max new names
** are nset to 0"

open-subshare/REPWRITERSPECS/BANNO.NEWSUBCREATE.V1.POW
```diff
@@ -546,34 +764,77 @@ PRINT TITLE="BANNO.NEWSUBCREATE.V1.POW"
...
+** allowed flag to 0
+*]
        PRINT "    "+Q+"canAddNames"+Q+": "
-       IF VALUE(SACNAMEALLOWFLAG)=0 THEN
+       IF VALUE(SACNAMEALLOWFLAG)=0 OR
+         (VALUE(SACNAMEALLOWFLAG)=1 AND
+          SACNAMEMAXNEW=0 AND
+          COPYNAMELISTCOUNT=0) THEN
```
jordan-keenan, line 780: Hi Tom, Won't COPYNAMELISTCOUNT still be 0 at this point in the program as BUILDCOPYNAMELIST hasn't been called yet, so the condition will fail in the RGSTATE NAMEPRELOAD? 
Or will Procedure BUILDCOPYNAMELIST, have already been called at this point in RGSTATE CREATESHARE? I can't tell looking at the PO code, as the UX code calls that.

open-subshare/REPWRITERSPECS/BANNO.NEWSUBCREATE.V1.POW
```diff
@@ -2525,6 +3183,403 @@ PROCEDURE PRINTPROGRAMINFO
...
+    DO
+     IF ADDNAMESSN=COPYNAMELIST(CNLSSN,TMPLOOP) THEN
+      DO
+       DUPLICATENAMEFOUND=TRUE
+       COPYNAMELIST(CNLDUPLICATED,TMPLOOP)="TRUE"
+      END
+    END
+  END
+
+ BNODEBUGMSG=BNODEBUGMSG+"DUP: "+
```
jordan-keenan, line 3361: I think you might want a space before the word DUP for the line with BNODEBUGMSG=BNODEBUGMSG+"DUP: "+

open-subshare/REPWRITERSPECS/BANNO.NEWSUBCREATE.V1.POW
```diff
@@ -323,22 +330,219 @@ DEFINE
...
+ CNLHOMEPHONE              = 29
+ CNLCARRIERROUTE           = 30
+ CNLCOUNTRYCODE            = 31
+ CNLCOUNTRY                = 32
+ CNLRESTRICT               = 33
+ CNLLASTADDRVERIFDATE      = 34
+ CNLSEX                    = 35
+ CNLDUPLICATED             = 36
+
+ CNTYPE                    = NUMBER
```
jordan-keenan, line 513: I don't see CNTYPE used in the code. If you don't plan to used it in the future, I suggest removing it.

open-subshare/REPWRITERSPECS/BANNO.NEWSUBCREATE.V1.POW
```diff
@@ -323,22 +330,219 @@ DEFINE
...
  ELSE
   DATAFILETYPE="DATA"
 
  Q=CTRLCHR(34)
- PROGRAMVERSION="1.0.9"
- LASTMODDATE='05/12/21'
- LASTMODTIME="16:30 PT"
- PROGRAMUPDATENOTE1="Corrected JSon output when no shares have and available balance"
- PROGRAMUPDATENOTE2="Enhanced Program info to JSON Output"   
+ QQ=CTRLCHR(34)
```
jordan-keenan, line 540: I also am not sure if we need two CTRLCHR(34) variables, unless sometimes QQ makes the code more readable?

open-subshare/REPWRITERSPECS/BANNO.NEWSUBCREATE.V1.POW
```diff
@@ -2416,46 +3072,44 @@ PROCEDURE ERRORHANDLER
...
+   ERRORDETAIL=ERRORDETAIL+ERRORCODEDESCR(ECDEFAULT,TMPNUM)
+
+   IF VALUE(ERRORCODE)=501 THEN
+    ERRORDETAIL=ERRORDETAIL+" - "+CONFIGFILENAME+": "+CFGFILEREADERROR
+   ELSE IF VALUE(ERRORCODE)=502 THEN
+    ERRORDETAIL=ERRORDETAIL+" - "+FORMAT("warning code 999",I)
+   ELSE IF VALUE(ERRORCODE)=503 THEN
+    ERRORDETAIL=ERRORDETAIL+" - "+FORMAT("Acct type 9999",ACCOUNT:TYPE)
+   ELSE IF VALUE(ERRORCODE)=505 THEN
+    ERRORDETAIL=ERRORDETAIL+" - "+FORMAT("Share Grp 99 ",VALUE(OPENCATEGORY))+LFERROR
```
jordan-keenan, line 3094: I think this error was more explicit before when it had 'term file:"+TMPCHR', consider adding it back unless you think that was redundant or would make the error too long.

jordan-keenan (changes requested):
Hi Tom,
 First, these changes are really good, adding much functionality to the Code.
I like that you even went the extra step of supporting comments in the terms letter files, and now the comments get to stay in the terms letter files as comments. 
And that is a nice touch sending an email to whatever person is maintaining the PO at the CU,
about copying existing share/loan names that have been added that had a duplicate SSN to another name, but the address is different.

Please consider each suggestion I that put in the code review as possible changes, before merging.

Thinking about it now, I think this comment I wrote will require a change, as while it might not be an issue during testing,
It could be an issue if two or more members are using the PO at the same time in Banno (not sure if you meant another parameter or something) .
“  For     IF VALUE(SACNAMEALLOWFLAG)=0 OR
         (VALUE(SACNAMEALLOWFLAG)=1 AND
          SACNAMEMAXNEW=0 AND
          COPYNAMELISTCOUNT=0) THEN
Hi Tom, Won't COPYNAMELISTCOUNT still be 0 at this point in the program as BUILDCOPYNAMELIST hasn't been called yet, so the condition will fail in the RGSTATE NAMEPRELOAD?
Or will Procedure BUILDCOPYNAMELIST, have already been called at this point in RGSTATE CREATESHARE? I can't tell looking at the PO code, as the UX code calls that.
“

I also added these comments in a nonstandard place in git hub for code reviews,
so I’m rewriting them in my review here.

In the config worksheet I suggest turning section ” Copying Existing Names” into 4 sections.
Perhaps,
“Copy Existing Share Names”,
“Copy Existing Share Name Valid Types List”,
“Copy Existing Loan Names” and
“Copy Existing Loan Name Valid Types List” ?

In the config worksheet in the Parameter description section of ”Error display messages”,
I suggest mentioning,
Configuration uses defaults values of:
Error condition 500: System is in Memo Post Mode
Error condition 501: Config file read error
Error condition 502: Account warning exists
Error condition 503: Ineligible account type
Error condition 504: No eligible shares to be created
Error condition 505: Error reading ToC for group
Error condition 506: No Shares with sufficient funds to xfer
Error condition 507: Error reading passed name info
Error condition 508: Next ID calculation error
Error condition 509: Error getting new share rate
Error condition 510: Could not calculate new maturity date
Error condition 511: Error creating new share

In the config worksheet in each Parameter description section ”Share Group ## Terms & Conditions”,
I had made a mistake omitting the line “Avoid the use of double quotes”. Sorry.
So I recommend for each group changing:
“Of course... don't forget to include them in their required delimiters '<>'.
or special characters in your message.”
To instead say:
“Of course... don't forget to include them in their required delimiters '<>'.
Avoid the use of special characters in your message.”
Or you could add you new improved comments instead for that Parameter Description.

jordan-keenan (approved):
The code changes look great!
I agree with Justin; especially the improvements you made to the configuration worksheet look very good!

PR #143, opened 2022-10-13
BANNO.CDRENEW.V1.CFG change term option update

tkainz:
Added clarifying description for option #2  - change cert term

### DIFF
--- a/certificate-renewal/LETTERSPECS/BANNO.CDRENEW.V1.CFG
+++ b/certificate-renewal/LETTERSPECS/BANNO.CDRENEW.V1.CFG
@@ -17,6 +17,8 @@
 *             share type
 * 05/11/2022: RRobison - Banno
 *             Added support for eligible share IRS codes
+* 10/13/2022: TKainz - Banno
+*             Added clarification for renewal Option 2
 *
 * PROGRAM PARAMETER SETTINGS START HERE ********************************
 *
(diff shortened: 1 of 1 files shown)


certificate-renewal/LETTERSPECS/BANNO.CDRENEW.V1.CFG
```diff
@@ -31,6 +33,11 @@
 *   OPTION4: Renew certificate (roll over)
 *   OPTION5: Disburse funds by check
 *   OPTION6: Suspend certificate
+*
+*   OPTION 2, if selected by the member, it results in the member being
```
jucarson, line 37: Looks good to me, I would just remove the word "it" on line 37 or correct the grammar in another way.

tkainz: Yep, I would agree.... but I was asked to specifically add it in so..?   :-)

jucarson: alright! :)

jucarson (approved):

certificate-renewal/LETTERSPECS/BANNO.CDRENEW.V1.CFG
jordan-keenan: Sorry I got behind. 
Yes; the change looks good 
to let the person setting up the PO config know what Option 2 does,  
as the UX just tells the member to contact the CU to change the certificate term when they select that in Banno.

PR #147, opened 2022-10-31
Banno NewSub Create Config Version Chect

tkainz:
Added code to CONFIG program to add version number and version date to CFG file.  Added code to PowerOn program to read and test for valid configuration file. Also corrected some config file read error variable names.


open-subshare/REPWRITERSPECS/BANNO.NEWSUBCREATE.V1.CONFIG
```diff
@@ -125,13 +125,20 @@ DEFINE
...
 SETUP
 [* Set default values
 *]
  DEBUGMODE=DEBUGMODEOFF
  Q=CTRLCHR(34)
+[* Configuration version must be in the following format:
+**   'a.b.c mm/dd/yyyy'
+**      a=major ver, b=minor ver, c=patch
+*]
+ CONFIGPROGRAMVER="1.1.0 10/20/2022"
```
rrobison44, line 141: Maybe move the config file information to the top of the setup devision.  It should probably be the first thing defined in the setup.

open-subshare/REPWRITERSPECS/BANNO.NEWSUBCREATE.V1.POW
```diff
@@ -517,6 +519,7 @@ SETUP
  TRUEFALSECHR(FALSE)="false"
  TRUEFALSECHR(TRUE)="true"
  SACSHAREGROUPSMAX=19
+ CONFIGPROGRAMMINDATE='10/20/22'
```
rrobison44, line 522: Maybe group this version information with the other version variables below.  And maybe have all the version information as the first thing in the setup division.

Group CONFIGPROGRAMMINDATE with these: 
`PROGRAMVERSION="1.1.0"
 LASTMODDATE='10/20/22'
 LASTMODTIME="12:00 MT"
 PROGRAMUPDATENOTE1="Copy name functionality added"
 PROGRAMUPDATENOTE2="Error code output update, JSON code fix"`

rrobison44 (changes requested):
Looks good.  Added a couple suggestions.

open-subshare/REPWRITERSPECS/BANNO.NEWSUBCREATE.V1.CONFIG
tkainz: Moved.  Added additional comments and moved a couple other items (debug mode setting and quote ctrl value to the init procedure.

open-subshare/REPWRITERSPECS/BANNO.NEWSUBCREATE.V1.POW
tkainz: Done

rrobison44 (approved):

jucarson (approved):
Looks good!

jordan-keenan (commented):
The code looks good.
I'm glad there is a way now to make sure the  CU is using the right configuration before they get the V3 UX.

However, I suggest the Variable CONFIGPROGRAMMINDATE
be changed to CONFIGPROGRAMMINGDATE, probably change in this code, as we
will tend to copy this code for future Banno PowerOn Config Versioning.
Otherwise MINDATE sounds like Minimum date (unless that is what you mean).

Sorry if you do mean Minimum date, I was at a Dr. Appointment during the
Thursday meeting on October 27th.

tkainz:
> The code looks good. I'm glad there is a way now to make sure the CU is using the right configuration before they get the V3 UX.
> 
> However, I suggest the Variable CONFIGPROGRAMMINDATE be changed to CONFIGPROGRAMMINGDATE, probably change in this code, as we will tend to copy this code for future Banno PowerOn Config Versioning. Otherwise MINDATE sounds like Minimum date (unless that is what you mean).
> 
> Sorry if you do mean Minimum date, I was at a Dr. Appointment during the Thursday meeting on October 27th.

@jordan-keenan the MINDATE represents the minimum date the config program last mod date must be. As for renaming the variable let's leave it as is for the moment.... this scenario is a little unique as the newsub create program is the only one with it's dedicated configuration program.  When we create a similar procedure for all of the other programs which utilize a configuration Letter file, we can be more critical on the variable names.

PR #151, opened 2022-12-05
Adding BANNO.GETTRACKING.V1.POW PowerOn to GH

rrobison44:
Adding BANNO.GETTRACKING.V1.POW to GitHub


jucarson (approved):
Looks good to me! Wish there were a way to grab all the data in loops but it is what it is! :)

PR #154, opened 2023-03-02
fix to subshare config and test account input

rrobison44:
With more than 5 test accounts the <input> line exceeded the 132 char maximum resulting in invalid html.  Trying to add a new test account would fail -- likely due to the broken html on the input.

### DIFF
--- a/open-subshare/REPWRITERSPECS/BANNO.NEWSUBCREATE.V1.CONFIG
+++ b/open-subshare/REPWRITERSPECS/BANNO.NEWSUBCREATE.V1.CONFIG
@@ -37,6 +37,9 @@
 **              specfile.
 **              2. New fee support via BANNO.NEWSUBCREATE.FEES.V1.POW
 **              3. Added support for three new error codes (512, 513 and 514).
+**  Ver. 1.2.1  03/02/2023: RRobison
+**              Fixed bug with test account input.  Code was not saving more than 6 test accounts
+**              when entered in the Test Accounts field.
 **
 **  This Banno service PowerOn is to be installed for use on
 **  demand and is run from the Account Manager workspace to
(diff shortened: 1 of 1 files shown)


open-subshare/REPWRITERSPECS/BANNO.NEWSUBCREATE.V1.CONFIG
```diff
@@ -1176,8 +1179,8 @@ PROCEDURE DISPLAYHTML
 
  HTMLVIEWLINE("<td class='lbl width300'>Test accounts list:</td>")
  HTMLVIEWLINE("<td>")
- HTMLVIEWLINE("<input id='testModeAccountList' class='width600' type='text' value='"+
-  SACTESTMODEACCOUNTLIST+"' ")
+ HTMLVIEWLINE("<input id='testModeAccountList' class='width600' type='text'")
```
tkainz, line 1182: This fix is good as it currently is but just to be safe, should we add a maxlength='109' to the input to limit the user input to 10 account numbers and 9 delimiters between them?

tkainz (approved):
added additional suggestion to fix, but not required.

open-subshare/REPWRITERSPECS/BANNO.NEWSUBCREATE.V1.CONFIG
tkainz: Also.... have you checked the other input fields to make sure there are no similar potential issues?

jordan-keenan (approved):
Code Fix looks good!
I think Tom's Idea is also good, if you have time to implement that.

open-subshare/REPWRITERSPECS/BANNO.NEWSUBCREATE.V1.CONFIG
rrobison44: added maxlength

PR #156, opened 2023-04-12
Sub Acct Rate Calc Updates

jucarson:
Sub Acct Rate Calc Updates:

- Corrected rate calcs for balance tables starting with a balance limit of $0.00.
- Corrected rate calc issue where 10th balance range in table was not being read.


tkainz (approved):

jordan-keenan (approved):
Changes look good!

PR #160, opened 2023-06-08
PREFACCESS.SSNMATCHING.BATCH ReadMe

tkainz:
Added ReadMe file


jordan-keenan (changes requested):
The Read Me Looks very good Tom!
I just have this finding.

For "Valid name types."
 It says "Target account name types which the program will ignore."
I think you mean something like "Valid Target account name types which the program checks for."

tkainz:
Updated document per Jordan's review

jordan-keenan (approved):
The changes look good!

PR #168, opened 2023-09-20
Update BANNO.CHANGE.ADDR.V1.POW

jucarson:
Member note fixes


change-address/REPWRITERSPECS/BANNO.CHANGE.ADDR.V1.POW
```diff
@@ -1075,7 +1076,7 @@ PROCEDURE PULLRGDATA
    PARSEDTEXTSTART=2
    PARSETEXTLENGTH=40
    CALL PARSETEXTFIELD
-   BNODEBUGMSG="RG3:"+SEGMENT(@RGUSERCHR3,1,110)
+   BNODEBUGMSG="RG3:"+SEGMENT(@RGUSERCHR3,1,130)
```
tkainz, line 1079: I believe this may cause the JSON output to exceed 132 characters and thus hose your whole JSON formatting.  "RG3:"+SEGMENT(@RGUSERCHR3,1,130) is already 134 characters - not to mention the quotes added to the debug output for each line. You may want to keep this at 110 for the debug output only as that should be sufficient to assist any debugging process.

tkainz (commented):
Double check what your debug output line length "could" be with this change.

change-address/REPWRITERSPECS/BANNO.CHANGE.ADDR.V1.POW
jucarson: Thanks Tom, you're 100% right. Also, I found 2 other spots where that was happening and fixed it there too. Good catch.

tkainz (approved):
Thanks!!

PR #171, opened 2023-11-01
Cus 2744 check wd soft text for certain error codes

rrobison44:
Add the ability to set custom error text for a subset of error codes.  If the `displayErrorMessage' property is included in an error response, the UX will display that error message instead of what is hard-coded in the UX.  The PO supports this feature for the following error codes:
- 501 - Insufficient funds
- 502 - Missing/invalid address
- 503 - Share/loan ineligible for check withdrawal 
- 504 - RegD limit
- 505 - Cross account withdrawals not allowed

[Jira CUS-2744](https://banno-jha.atlassian.net/browse/CUS-2744)


withdraw-by-check/withdraw-by-check-json-contract.md
```diff
@@ -48,76 +53,90 @@
...
+- owner: The primary member's long name (NAME:LONGNAME)
+- address: The system calculated mailing address (ACCOUNT:PAYEELINE[1-6])
+- disclaimerText: Custom terms and conditions as set up in the parameter settings Letter file.
 
 **When error condition is encountered:**
+
 ```json
 {
   "errorCode": 999,
   "loggingErrorMessage": "error message detail"
```
jordan-keenan, line 81: Comma needed after "loggingErrorMessage": "error message detail"

withdraw-by-check/withdraw-by-check-json-contract.md
```diff
@@ -126,11 +145,14 @@ Unless otherwise noted, all values will be passed as double-quote encapsulated c
...
- - memoMode: Is the system in MemoMode? (boolean)
+
+- success: Was the attempt successful? (boolean)
+- memoMode: Is the system in MemoMode? (boolean)
 
 **When error condition is encountered:**
+
 ```json
 {
   "errorCode": 999,
```
jordan-keenan, line 158: this new line is needed "displayErrorMessage": ["display error message detail"] , right after Line "loggingErrorMessage": "error message detail",

withdraw-by-check/withdraw-by-check-json-contract.md
rrobison44: Fixed.

withdraw-by-check/withdraw-by-check-json-contract.md
rrobison44: Fixed.

withdraw-by-check/REPWRITERSPECS/BANNO.CHECK.WITHDRAW.V1.POW
```diff
@@ -175,374 +266,357 @@ PRINT TITLE="BANNO.CHECK.WITHDRAW.V1.POW"
...
+   PARAMERRDISPLAY(TMPLOOP,ERRDISPLAYERRCODE)=0
+   PARAMERRDISPLAY(TMPLOOP,ERRDISPLAYLINECOUNT)=0
   END
 
- PRINT "]"
- PRINT "  }"
- NEWLINE
+ VALIDATESTATUS=VALIDATEOK
+ TRANSTATUS=TRANOK
+
```
jordan-keenan, line 393: The following was removed from INITAILIZEDATA. I think it should be added back:

 FOR TMPLOOP=0 TO PARAMCUSTOMTERMLINESMAX
  DO
   PARAMCUSTOMTERMS(TMPLOOP)=""
  END
 PARAMCUSTOMTERMLINES=0

withdraw-by-check/REPWRITERSPECS/BANNO.CHECK.WITHDRAW.V1.POW
```diff
@@ -1061,5 +1166,387 @@ PROCEDURE BNODEBUGMSGADD
...
+   JSONPROPNAME="maxWdAmount"
+   CALL JSONPROP
+   PRINT Q+TMPCHR+Q
+   NEWLINE
+  END 
+END [PROCEDURE]
+
+[*
+**  --------------------------------------------------------------------
+**  JSON Procedures - Common
```
tkainz, line 1400: You have a few procedures in this section of the program (JSON procedures) which just have one line of code (eg: PROCEDURE JSONOBJ). Does this add unnecessary stress on the system to divert program run to a procedure to run one line of code... instead of just running that one line of code from the original location?

withdraw-by-check/REPWRITERSPECS/BANNO.CHECK.WITHDRAW.V1.POW
```diff
@@ -1061,5 +1166,387 @@ PROCEDURE BNODEBUGMSGADD
...
+
+PROCEDURE NLS
+[* Remove leading spaces
+**  PASS       TMPCHR
+**  RETURN     TMPCHR
+*]
+ WHILE (LENGTH(TMPCHR)>0 AND
+        SEGMENT(TMPCHR,1,1)=" ")
+  DO
+   IF LENGTH(TMPCHR)=1 THEN
```
tkainz, line 1494: I know this is old code.......but as used here... IF LENGTH(TMPCHR)=1 can never be.... since in order to enter the DO/END structure, the SEGMENT(TMPCHR,1,1) must equal=" " and since LENGHT ignores trailing spaces, TMPCHR can never be " " and have a length of 1 at the same time.

Maybe....

```
 WHILE (LENGTH(TMPCHR)>0 AND
        SEGMENT(TMPCHR,1,1)=" ")
  DO
   IF TMPCHR=" " THEN
    TMPCHR=""
   ELSE
```

withdraw-by-check/REPWRITERSPECS/BANNO.CHECK.WITHDRAW.V1.POW
```diff
@@ -1061,5 +1166,387 @@ PROCEDURE BNODEBUGMSGADD
...
+ IF LENGTH(TMPCHR)=0 THEN
+  TMPCHR=""
+END [PROCEDURE]
+
+PROCEDURE NTS
+[* Remove trailing spaces
+**  PASS       TMPCHR
+**  RETURN     TMPCHR
+*]
+ IF TMPCHR<>"" THEN
```
tkainz, line 1508: Old code again... can't this procedure just be reduced to 1 line?

```
TMPCHR=SEGMENT(TMPCHR,1,LENGTH(TMPCHR))
```

withdraw-by-check/REPWRITERSPECS/BANNO.CHECK.WITHDRAW.V1.POW
```diff
@@ -706,6 +805,25 @@ PROCEDURE READCONFIGFILESETTINGS
            PARAMCUSTOMTERMS(PARAMCUSTOMTERMLINES)=SEGMENT(TMPCHR,1,123)
           END
         END
+
+[* Custom error display messages
+*]
+       ELSE IF SEGMENT(PARAMNAME,1,3)="EDM" THEN
+        DO
+         TMPERRCODE=VALUE(SEGMENT(PARAMNAME,4,6))
```
tkainz, line 813: Just a consideration... Instead of looking for specific segment here to get the error code value, maybe consider using the VALUE statement on the full PARAMNAME to eliminate the possibility that the user accidently includes a space, or adds a dash, etc. in between the 'EDM' and the error code.

tkainz (commented):
Some minor considerations

withdraw-by-check/REPWRITERSPECS/BANNO.CHECK.WITHDRAW.V1.POW
rrobison44: By setting PARAMCUSTOMTERMLINES to zero, you are initializing the array.  I don't see a need to set each element in the array to a blank screen.  We should only be accessing things from 0 to PARAMCUSTOMTERMLINES in the array.  Let me know if you see another reason why that extra code should exist.

withdraw-by-check/REPWRITERSPECS/BANNO.CHECK.WITHDRAW.V1.POW
```diff
@@ -797,174 +915,160 @@ PROCEDURE CONFIGFILEPOSTPROCESS
...
-      DO
-[* Check for back-slash or double-quote. If found, precede with a back-slash
-*]
-       IF (CHRVAL=92 OR CHRVAL=34) THEN
-        TMPTEXT=TMPTEXT+"\"+SEGMENT(TMPCHR,TMPTEXTLOOP,TMPTEXTLOOP)
-       ELSE
-        TMPTEXT=TMPTEXT+SEGMENT(TMPCHR,TMPTEXTLOOP,TMPTEXTLOOP)
-      END
+     BNOERRFOUND=TRUE
+     IF BNOERRMESSAGE<>"" THEN
```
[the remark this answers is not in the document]
rrobison44: Agree.  Originally, that procedure could also update the BNOERRDISPLAYMSG.  That is why I had that because there could have been cases where BNOERRMESSAGE was not being updated.  Fixed.

withdraw-by-check/REPWRITERSPECS/BANNO.CHECK.WITHDRAW.V1.POW
```diff
@@ -134,21 +199,47 @@ DEFINE
...
  BNOPRINTDEBUGINFODAYS=90
 
+ JSONBOOLEAN(TRUE)="true"
+ JSONBOOLEAN(FALSE)="false"
+
+ BNOERRCOUNT=0
+ BNOERRMESSAGE=""
+ BNOERRDISPLAYMSGCOUNT=0
+
+ BNOERRCODE=BNOERRCODECFGPARSE          BNOERRMESSAGE="Error reading config file"
```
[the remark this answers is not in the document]
rrobison44: Because this PO had a bunch of cases where a single ERRORCODE actually had multiple different text strings for the same BNOERRCODE.  So the initialization just sets the baseline text.  But then, in certain cases, the message gets overriden with more specific text.

withdraw-by-check/REPWRITERSPECS/BANNO.CHECK.WITHDRAW.V1.POW
```diff
@@ -1061,5 +1166,387 @@ PROCEDURE BNODEBUGMSGADD
...
+     NEWLINE
+    END
+   CALL JSONARRAYEND
+  END
+
+ IF TRANSTATUS=TRANLIMITS AND
+    BNOERRCODE=BNOERRCODEINVALIDAMT THEN
+  DO
+   PRINT ","
+   TMPCHR=FORMAT(" ###,##9.99",RGWDAMOUNT)
```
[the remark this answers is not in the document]
rrobison44: I don't know.  This was existing.  Is that a placeholder for a negative sign?  But I did check all the places where the comma is included and it is only used in the error scenario and would be part of a string that is in the json -- which would be fine.  You would only see it if you checked the payload using dev tools.  The actual money fields (like minWdAmount) don't include the comma or a space.

withdraw-by-check/REPWRITERSPECS/BANNO.CHECK.WITHDRAW.V1.POW
rrobison44: I would not think so.  I think the benefit outweighs any concern in those regards.  Using those procedures just standardizes the way json objects, arrays, etc. are opened and closed and prevents typos.  That's my take...

withdraw-by-check/REPWRITERSPECS/BANNO.CHECK.WITHDRAW.V1.POW
rrobison44: I feel like the the correct format is detailed in the CFG and probably should not be interpreting the param differently than the required/specific format.  
```
Line must be formatted as "EDMxxx:" where xxx represents the 3-digit error code that is returned to the UX.
```
I get what you are saying but how would something like "EDM  5     5FDFD  6". Would that actually be valid in that we would interpret it as error code 556?

jordan-keenan (approved):
Changes look good! Thank you

withdraw-by-check/REPWRITERSPECS/BANNO.CHECK.WITHDRAW.V1.POW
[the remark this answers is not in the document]
tkainz: OK

withdraw-by-check/REPWRITERSPECS/BANNO.CHECK.WITHDRAW.V1.POW
tkainz: OK

withdraw-by-check/REPWRITERSPECS/BANNO.CHECK.WITHDRAW.V1.POW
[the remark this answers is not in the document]
tkainz: OK

withdraw-by-check/REPWRITERSPECS/BANNO.CHECK.WITHDRAW.V1.POW
tkainz: works for me then

tkainz (approved):
Thank you!

jucarson (approved):
Looks good!

PR #176, opened 2023-12-11
Update BANNO.CHANGE.ADDR.V1.POW

jucarson:
Add new address information to account level note during memo mode

### DIFF
--- a/change-address/REPWRITERSPECS/BANNO.CHANGE.ADDR.V1.POW
+++ b/change-address/REPWRITERSPECS/BANNO.CHANGE.ADDR.V1.POW
@@ -32,8 +32,8 @@
 **                                   - Added validation to ensure street, city, state and zip are not blank.
 **    Ver. 1.2.1  09/20/23 J. Carson - Increased size of member note from 110 characters to 130. Fixed issue 
 **                                     with truncating long member notes.
-**    Ver. 1.2.2  12/11/23 J. Carson - Added member's input address to account level note when host is in 
-**                                     memo mode. 
+**    Ver. 1.2.2  12/13/23 J. Carson - Added member's input address to account level note when host is in 
+**                                     memo mode or a config file error occurs. 
 **
 *]
 SYMCONNECT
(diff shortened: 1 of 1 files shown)


rrobison44 (approved):

change-address/REPWRITERSPECS/BANNO.CHANGE.ADDR.V1.POW
```diff
@@ -253,10 +255,10 @@ PRINT TITLE="BANNO.CHANGE.ADDR.V1.POW"
   CALL READCONFIGFILESETTINGS
 
  IF RESPONSECODE=0 THEN
-  CALL CHECKFORMEMOMODE
+  CALL PULLRGDATA
```
jordan-keenan, line 258: Does this code only add the member's input address to an account level note if it would make the error "Address entry exceeds Episys limits" but Episys is in Memo mode?

Looking the code it won't make the note if there is no error "Address entry exceeds Episys limits".
Is that OK for it not to make the account level note in that situation per the requirements?

jucarson: Hey Jordan, Can you explain that for me? I do see that the check for too many characters is the last thing the the PULLRGDATA procedure does. So if they do exceed limits, the will get that error only regardless of memo mode status. If they don't exceed limits, then the RESPONSECODE is still 0. If it is 0, then it will check for memo mode and it will give that error only. Then, if it is still 0, it proceeds on down the line.

jordan-keenan: Yes; That was my question. Is the point just to make sure the note about the address the member attempted change being stored in a note on their account whenever it would have been an error for "Address entry exceeds Episys limits", even if it is in Memo Mode. If that is the case then this code looks good to me.  
  If the point is to have a note put in at the account level whenever a member changes the address even if the program is in memo mode, I think that won't work as if there was no error and RESPONSECODE is 0, and then for Memo Mode check RESPONSECODE won't be 0.

jordan-keenan: It really a problem. But thank you for making a new change and making the code even better.

jordan-keenan (approved):
Thank you for going the extra mile.
Code changes look good!

tkainz (approved):
Makes Sense!

rrobison44 (approved):
Latest changes look good.

PR #186, opened 2024-03-21
Updated Configuration Worksheet

ebeltinck:


tkainz (changes requested):
I realize that portions of the CGF Worksheet were copied from the program header but in reading it through, I think additional attention should be given to the "Cross Account Access" parameter setting. As it is now, I feel it doesn't seem to convey enough information to the end user accurately assess it's impact..  Maybe add some of the detail contained in the actual procedure header:
```
**    'CROSS ACCOUNT ACCESS' (as XA:)
**       This setting is used to determine which accounts will be
**       searched through looking for the target transactions/transfers.
**       If cross account access is turned on then this program will look
**       through all accounts where the account activity date is on or after
**       the reporting start date, otherwise, the focus will be narrowed down
**       to only those accounts with an HB user who has last logged into
**       home banking on or after the reporting start date.
```

eamadden:
Tom, is this regarding updates to the M2M Transfer Configuration Worksheet?  We are still working on that one, so I will make note of that when we get to finalizing that document and compare to what you are using for the CFG param description.  Thank you!

PR #189, opened 2024-05-13
BANNO.LOANPAYMENT.SKIP.V1.POW Install Date Param

tkainz:
Added installation date parameter.


jucarson (approved):
Looks Good!

loan-payment-skip/REPWRITERSPECS/BANNO.LOANPAYMENT.SKIP.V1.POW
```diff
@@ -1106,6 +1111,20 @@ PROCEDURE READCONFIGFILESETTINGS
             VALIDTESTACCOUNT=TRUE
           END
         END
+
+[ get program installation date ]
+       ELSE IF TMPPARAM="PID" AND
```
rrobison44, line 1116: In this block of code should we be setting BNODEBUGMODE to BNODEBUGMODEOFF if we find the param but we are beyond the 90 days?  As written. I think we fall back to the old way of comparing against the last mod date.  I think, to be in sync with how it is described in the CFG, we should set BNODEBUGMODE to BNODEBUGMODEOFF when the param is found but we are beyond the 90 days.  I also noticed there is no initialization on the BNODEBUGMODE variable.  It probably gets set to BNODEBUGMODEOFF automatically but it might not be a bad idea to add the initialization just for clarity.

rrobison44 (changes requested):
See comments.

loan-payment-skip/REPWRITERSPECS/BANNO.LOANPAYMENT.SKIP.V1.POW
tkainz: @rrobison44: corrected - BNODEBUGMODE initialized and BNODEBUGMODE remains off if cutoff date (PID + BNOPRINTDEBUGINFODAYS) has been met/passed.

rrobison44 (approved):
Changes look good.

loan-payment-skip/LETTERSPECS/BANNO.LOANPAYMENT.SKIP.CFG
```diff
@@ -207,7 +209,17 @@ DQE:0
...
+*  Enter the date the program was activated in the production system.
+*  Additional program data will be accessible for 90 days after the
+*  installation date to assist with issues tracking.
+*  VALID VALUES: Any valid date on or prior to the system date. in the
+*                following format: mm/dd/yyyy.
+*  DEFAULT: In absence of a date, 90 days from the program's most recent
+*           modification date will be used.
+PID:
 * PROGRAM TIME PERIOD
 * The calender date range during which this skip program is available to
```
jordan-keenan, line 224: @tkainz Code Looks Good!
I would just change the spelling of "calender" in the CFG on line 224 where is says "* The calender date range during"

jordan-keenan (approved):
@tkainz Code Looks Good!
I would just change the spelling of "calender" in the CFG on line 224 where is says "* The calender date range during"

PR #191, opened 2024-05-14
BANNO.NEWSUBCREATE.V1.POW program install date param

tkainz:
Added program installation date parameter and JSON output. cleaned up trailing spaces.


jordan-keenan (commented):
Hi @tkainz, I thought PARAMINSTALLDATE was going to be a configuration file level setting.
It seems like it would make more sense to have the person installing the Program Either fill out the date or forget to fill out the date when they are setting up the configurations, rather than have them look an additional place in the PowerOn.

That's just my two cents worth

tkainz:
@jordan-keenan I was thinking about that - whether or not to put it into the on-demand configuration program since the CFG data file is not accessible. Then, on the other side, I figured that since this is really meant to be a one-time entry - by either the CU or the BA installing the program, did we want to/need to add that code to that on-demand program.  Maybe @jucarson and @rrobison44 can chime in with their thoughts?

open-subshare/REPWRITERSPECS/BANNO.NEWSUBCREATE.V1.POW
```diff
@@ -560,13 +562,18 @@ DEFINE
...
 SETUP
- PROGRAMVERSION="1.3.1"
- LASTMODDATE='03/28/24'
- LASTMODTIME="15:00 MT"
- PROGRAMUPDATENOTE1="malformed json fix"
+ PROGRAMVERSION="1.4.0"
+ LASTMODDATE='05/14/24'
+ LASTMODTIME="12:00 MT"
+
+ PARAMINSTALLDATE='05/14/24'
```
rrobison44, line 574: Is this correct?  Shouldn't the code just be pulling the PARAMINSTALLDATE from the CFG?  OH....maybe because we have the CFG program for this one?  And that is not being updated?

rrobison44 (approved):
Approved but added a comment.

open-subshare/REPWRITERSPECS/BANNO.NEWSUBCREATE.V1.POW
rrobison44: I see this discussed in another thread...will respond there.

rrobison44:
> Good Morning @tkainz. I think for consistency's sake, the install date should be in the config datafile. This way there is no need to remember that, with this one program, the PO needs to be updated instead of the config file. My vote would be to update the .CONFIG PO with code to add the program installation date to the datafile when the datafile is created (not updated) with no user intervention required at all.
> 
> Just my 2 cents :)

I agree with Justin/Jordan on this.

tkainz:
@jucarson, @jordan, & @rrobison44: I've added the program installation date parameter to the CONFIG program and updated the POW accordingly. I also moved a lot of the initializations in the POW from the setup division into their own procedure, called from the print div.  Please re-review at your convenience.

Thanks!

jucarson (approved):
Looks good!

jordan-keenan (approved):
@tkainz Changes look good, Tom!

rrobison44 (approved):
Updates look good.

PR #193, opened 2024-05-30
Update BANNO.LOANPAYMENT.SKIP.V1.POW

jucarson:
CU reports that if a relationship code is set to no fee ($0.00), member is still charged base fee. I found the issue in the functionality that sets an unset relationship code to the base fee.


jordan-keenan (approved):
Changes look good! Thanks for doing this!

rrobison44 (approved):
Code looks good.  A suggestion would be to have constant like `FEEAMOUNTUNDEFINED=-$0.01` then initialize to `FEEAMOUNTUNDEFINED` and then check for `IF FEEAMOUNT=FEEAMOUNTUNDEFINED` when setting the fee amount to the default fee amount.  Doing something like that adds clarity in that we explicitly know it is undefined vs. having to know an amount of less than $0.00 is undefined.  Not a required change but something to consider.

PR #196, opened 2024-06-06
BANNO.CDRENEW.V1.POW add PID parameter

tkainz:
Added program installation date parameter

### DIFF
--- a/certificate-renewal/REPWRITERSPECS/BANNO.CDRENEW.V1.POW
+++ b/certificate-renewal/REPWRITERSPECS/BANNO.CDRENEW.V1.POW
@@ -77,7 +77,7 @@
 **                Added parameter setting to delete/expire any transfer type 3
 **                records pointing to the targeted certificate upon the member
 **                submitting their selection.
-**    Ver. 1.6.0  06/06/24 TKainz - Banno
+**    Ver. 1.5.1  06/06/24 TKainz - Banno
 **                         Added Program Installation Date parameter
 **
 **  DO NOT MODIFY THIS FILE UNLESS YOU KNOW WHAT YOU'RE DOING!
@@ -350,10 +350,10 @@ DEFINE
  SXFER3SHAREID             = CHARACTER(4) ARRAY(99)
  STRANSFERCOUNTMAX         = 99
 
-[* Added 06/06/24 *]
+[* TKainz - added 06/06/24
+*]
  PARAMINSTALLDATE          = DATE
  PARAMDEBUGMODE            = NUMBER
- PARAMDEBUGMODEFOUND       = NUMBER
 END [DEFINE]
 
 SETUP
@@ -362,8 +362,10 @@ SETUP
  BNOLASTMODTIME="12:00 PT"
  BNOPROGRAMNAME="BANNO.CDRENEW.V1.POW"
  BNOCFGFILENAME = "BANNO.CDRENEW.V1.CFG"
- BNOPROGRAMVERSION="1.6.0"
+ BNOPROGRAMVERSION="1.5.1"
  BNOPRINTDEBUGINFODAYS=90
+
+ BNODEBUGMODE=BNODEBUGMODEOFF
  IF BNOLASTMODDATE+BNOPRINTDEBUGINFODAYS>SYSTEMDATE THEN
   BNODEBUGMODE=BNODEBUGMODEON
 
@@ -741,7 +743,7 @@ PROCEDURE READCONFIGFILESETTINGS
 *]
  INVALIDCONFIGPARAMS=FALSE
  READCONFIGERROR=FALSE
- PARAMDEBUGMODEFOUND=FALSE
+ PARAMDEBUGMODE=-1
 
  FILEOPEN("LETTER",BNOCFGFILENAME,"READ",LFNUMBER,LFERROR)
  IF LFERROR<>"" THEN
@@ -961,12 +963,12 @@ PROCEDURE READCONFIGFILESETTINGS
 [* Debug mode. This is a hidden parameter setting which can override the default debug
 ** mode setting. Use this "DB:" parameter setting in the CFG Letter file with "ON"
 ** or "OFF" to force the debug mode to be other than the default value of being on for
-** the first 90 days after the most recent program modification date.
+** the first 90 days after the most recent program modification date orprogram installation
+** date, whichever is more recent
 *]
        ELSE IF PARAMNAME="DB" AND
                PARAMVALUE<>"" THEN
         DO
-         PARAMDEBUGMODEFOUND=TRUE
          IF CHARACTERSEARCH("ONTRUEYES",UPPERCASE(PARAMVALUE))>0 THEN
           PARAMDEBUGMODE=BNODEBUGMODEON
          ELSE IF CHARACTERSEARCH("OFFFALSENO",UPPERCASE(PARAMVALUE))>0 THEN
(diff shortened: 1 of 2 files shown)


certificate-renewal/REPWRITERSPECS/BANNO.CDRENEW.V1.POW
```diff
@@ -944,15 +952,27 @@ PROCEDURE READCONFIGFILESETTINGS
...
         DO
-         IF PARAMVALUE="TRUE" THEN
-          BNODEBUGMODE=TRUE
-         ELSE IF PARAMVALUE="FALSE" THEN
-          BNODEBUGMODE=FALSE
-       END
+         PARAMDEBUGMODEFOUND=TRUE
+         IF CHARACTERSEARCH("ONTRUEYES",UPPERCASE(PARAMVALUE))>0 THEN
+          PARAMDEBUGMODE=BNODEBUGMODEON
+         ELSE IF CHARACTERSEARCH("OFFFALSENO",UPPERCASE(PARAMVALUE))>0 THEN
```
rrobison44, line 972: Can we just get rid of this and assume it is off if it is not one of the valid ON values?  And then just make sure you are initializing PARAMDEBUGMODE.

rrobison44 (changes requested):
Added a few comments.  Might want to also initialize BNODEBUGMODE to OFF even though that probably happens by default.

certificate-renewal/REPWRITERSPECS/BANNO.CDRENEW.V1.POW
```diff
@@ -346,17 +348,21 @@ DEFINE
...
 
 SETUP
  INCLUDEPROGRAMINFO=TRUE
- BNOLASTMODDATE='03/14/24'
+ BNOLASTMODDATE='06/06/24'
  BNOLASTMODTIME="12:00 PT"
  BNOPROGRAMNAME="BANNO.CDRENEW.V1.POW"
  BNOCFGFILENAME = "BANNO.CDRENEW.V1.CFG"
- BNOPROGRAMVERSION="1.4.1"
+ BNOPROGRAMVERSION="1.6.0"
```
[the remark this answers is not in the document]
tkainz: @rrobison44 meant this to actually be a 1.5.0 instead of a 1.6.0.  But I can see considering this as a fix instead of a feature bump - even though nothing is broken per-se that this is fixing.  I'll change it to a 1.4.2.

certificate-renewal/REPWRITERSPECS/BANNO.CDRENEW.V1.POW
tkainz: @rrobison44 leaving this in actually also allows the 'hidden' DB parameter setting - if used - to force debug mode off. It just seemed logical that the DB parameter switch should also allow the CU to turn the debug mode off, overriding the standard 90 days default of it being on or overriding the Program Installation Date, forcing the debug mode on as well.  It also allows the developers or BA to easily turn the debug code off thus allowing them to just see the pertinent JSON code when reviewing it for info. Let me know if you still want it out.

certificate-renewal/REPWRITERSPECS/BANNO.CDRENEW.V1.POW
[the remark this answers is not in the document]
tkainz: @rrobison44 Actually.... it would be 1.5.1 as the last set of mods to the program were not correctly reflected in the setting of the BNOPROGRAMVERSION variable to 1.5.0 as it should have been:
```
**    Ver. 1.5.0  03/14/24 TKainz - Banno
**                Added parameter setting to delete/expire any transfer type 3
**                records pointing to the targeted certificate upon the member
**                submitting their selection.
```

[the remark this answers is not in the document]
rrobison44: Yeah, I guess just be consistent with the version bumps you did for the other programs for which you added this param.

certificate-renewal/REPWRITERSPECS/BANNO.CDRENEW.V1.POW
rrobison44: I think it works fine as is.  But since you already have PARAMDEBUGMODEFOUND, and you are setting it to TRUE if DB is found and it has a value, then what you are describing still works without having the ELSE as long as you are initializing PARAMDEBUGMODE to OFF.  If they have some rogue value in DB (like "disabled") then I think the code still functions as if we are overriding debug mode with the DB param config and setting it to OFF -- as that would be the default.  i.e. having DB set to some non-blank value (other than ON/TRUE/YES) effectively forces debug mode to OFF.  It doesn't necessarily have to be OFF/FALSE/NO.  But I am good with whatever you decide.  I just think you can get rid of some code and still do, logically, the exact same thing you are doing with the ELSE clause.  Of course it is also possible that my thinking is off :)

rrobison44 (approved):

certificate-renewal/REPWRITERSPECS/BANNO.CDRENEW.V1.POW
tkainz: @rrobison44 
**I'll certainly take another look at what I have.** 

My goal was to have the following functionality:
1. Debug mode defaults to being off
2. Determine if debug mode should be turned on based upon the last mod date and the program installation date values. Turn on if warranted.
3. Check for a 'DB:' value in the CFG Letter file
- If there is no DB value in the CFG Letter file, or a DB parameter setting with an invalid value (or no value at all), use the last mod date/installation date logic results.
- If there is a DB value with a valid YES/ON/TRUE value then override the current (last mod date/installation date) logic and force the debug mode on
- If there is a DB value with a valid NO/OFF/FALSE value then override the current logic and force the debug mode off.

I was trying to avoid the "unless I specifically turn it on, I mean for it to be off" logic.

I'll take a further look at it.

tkainz: @rrobison44 I know you already approved this but taking your suggestion, I was able to 'condense' the logic a bit further and remove a few lines of the code. The logic is still behaving as expected. Thanks for the nudge!

certificate-renewal/REPWRITERSPECS/BANNO.CDRENEW.V1.POW
```diff
@@ -346,18 +348,24 @@ DEFINE
  SXFER3EXPIREDATE          = DATE ARRAY(99)
  SXFER3LOC                 = NUMBER ARRAY(99)
  SXFER3SHAREID             = CHARACTER(4) ARRAY(99)
-
  STRANSFERCOUNTMAX         = 99
+
+[* TKainz - added 06/06/24
+*]
+ PARAMINSTALLDATE          = DATE
```
jordan-keenan, line 355: I noticed PARAMINSTALLDATE doesn't get initialized. You may want to consider initializing it.

certificate-renewal/REPWRITERSPECS/BANNO.CDRENEW.V1.POW
```diff
@@ -969,6 +991,19 @@ PROCEDURE READCONFIGFILESETTINGS
...
+**   If the "DB" parameter is set to false, then force debug
+**     mode to off (overrides all other debug mode logic settings).
+**   If the "DB" parameter is set to true or the installation date was found and
+**     less than 90 days prior, set the debug mode to true.
+*]
+ IF PARAMDEBUGMODE=BNODEBUGMODEOFF THEN
+  BNODEBUGMODE=BNODEBUGMODEOFF
+ ELSE IF PARAMDEBUGMODE=BNODEBUGMODEON OR
+     (PARAMINSTALLDATE<>DATENULL AND
+      PARAMINSTALLDATE+90>SYSTEMDATE) THEN
```
jordan-keenan, line 1004: at line 1004 I suggest using BNOPRINTDEBUGINFODAYS instead of 90 in "PARAMINSTALLDATE+90>SYSTEMDATE) THEN"

jordan-keenan (approved):
Hi Tom,
The code condensing looks good. Code Approved!

A couple suggestions:
.I noticed PARAMINSTALLDATE doesn't get initialized. You may want to consider initializing it.
.At line 1004 I suggest using BNOPRINTDEBUGINFODAYS instead of 90 in "PARAMINSTALLDATE+90>SYSTEMDATE) THEN".

certificate-renewal/REPWRITERSPECS/BANNO.CDRENEW.V1.POW
tkainz: Corrected

tkainz: @jordan-keenan Corrected

certificate-renewal/REPWRITERSPECS/BANNO.CDRENEW.V1.POW
tkainz: @jordan-keenan changed

jucarson (approved):
Changes look good.

jordan-keenan (approved):
Code changes look good! 
@tkainz Thanks for initializing PARAMINSTALLDATE and using parameter BNOPRINTDEBUGINFODAYS instead of 90 days.

PR #204, opened 2024-09-13
Update BANNO.CDRENEW.V1.POW

jucarson:
1.5.3 - Added check for backslashes and quotes on transfer to/from share descriptions.


jordan-keenan (approved):
@jucarson  , Code looks good!  

Only thing I saw was that on lint 663 there is not check to make sure TRANSFERLISTNAME(TRANSFERLISTCOUNT) is not blank before calling CHECKFORINVALIDCHARS, (not sure if possible but would be if the Share or Loan ID is not found). I'm not sure if this should even be changed as CHECKFORINVALIDCHARS would simply return blank. Just bringing it up to look at.

tkainz (approved):
Looks good to me - Thanks Justin

rrobison44 (approved):

PR #206, opened 2024-09-16
BANNO.LOANPAYMENT.SKIP.V1.POW Added max lifetime skips param

tkainz:
Added a parameter setting to allow the CU to set the max number of skips per lifetime of the loan. Corrected skips per year parameter setting to correctly honor the skips per year setting where a blank parameter setting will result in there being no per year limit.

### DIFF
--- a/loan-payment-skip/REPWRITERSPECS/BANNO.LOANPAYMENT.SKIP.V1.POW
+++ b/loan-payment-skip/REPWRITERSPECS/BANNO.LOANPAYMENT.SKIP.V1.POW
@@ -60,12 +60,14 @@
 **             Corrected issue with relationship level fee amounts
 **             when a fee amount is set to $0.00.
 **  Ver. 1.2.0 06/03/2024: TKainz - Banno
-**             Added Program Installation Date parameter
+**             Added Program Installation Date parameter.
 **  Ver. 1.2.1 07/10/2024 JKeenan - Banno
-**             Updated the JSON for ineligibilityReason to include 
+**             Updated the JSON for ineligibilityReason to include
 **             an additional line only viewable in logs, that shows
 **             the ineligible loan type, loan service codes,
 **             and account and loan warning code numbers.
+**  Ver. 1.3.0 09/17/2024: TKainz - Banno
+**             Added Max Skips per Loan Life parameter.
 **
 **  For more information check
 **  https://github.com/Banno/banno-powerons
@@ -95,6 +97,7 @@ DEFINE
  LOANSEASON          = NUMBER
  GRACEDAYS           = NUMBER
  SKIPCOUNT           = NUMBER
+ LOANLIFESKIPCOUNT   = NUMBER
  TEMPLOOP            = NUMBER
  TIMEPERIOD          = CHARACTER
  ERRORMSGUSER        = CHARACTER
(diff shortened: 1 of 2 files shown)


jucarson (approved):
Looks good!

loan-payment-skip/REPWRITERSPECS/BANNO.LOANPAYMENT.SKIP.V1.POW
```diff
@@ -204,7 +209,8 @@ DEFINE
 
  DEFAULTFEEAMOUNT          = $35.00
  DEFAULTLOANTRACKINGTYPE   = 77
- DEFAULTMAXSKIPS           = 2
+ DEFAULTMAXSKIPSYEAR       = 0
```
jordan-keenan, line 212: Hi @tkainz , I request you change line 212 "DEFAULTMAXSKIPSYEAR = 0" back to a default of 2, or change the cfg to say the default is 0.  
Other than that The code looks good and I think this is a a good idea to add max skips per life of loan as we came across it as an addition CU's made in the NetTeller version of the PowerOn.

jordan-keenan (changes requested):
Hi @tkainz , I request you change line 212 "DEFAULTMAXSKIPSYEAR = 0" back to a default of 2, or change the cfg to say the default is 0.  
Other than that The code looks good and I think this is a a good idea to add max skips per life of loan as we came across it as an addition CU's made in the NetTeller version of the PowerOn.

loan-payment-skip/REPWRITERSPECS/BANNO.LOANPAYMENT.SKIP.V1.POW
tkainz: @jordan-keenan Because of the discrepancy between the default of 2 and how the program is currently coded: The CFG "DEFAULT" value is 2, but the instructions state that a blank value (which would imply the default value) results in no limit. However, the code treated a blank value as the default of 2. The decision was made to leave the CFG Letter file parameter setting for 2 as a default in place, but set the code up such that a blank value in the CFG Letter file would indeed result in there being no limit.

jordan-keenan: > @jordan-keenan Because of the discrepancy between the default of 2 and how the program is currently coded: The CFG "DEFAULT" value is 2, but the instructions state that a blank value (which would imply the default value) results in no limit. However, the code treated a blank value as the default of 2. The decision was made to leave the CFG Letter file parameter setting for 2 as a default in place, but set the code up such that a blank value in the CFG Letter file would indeed result in there being no limit.

OK. Thanks for explaining, Tom. It's a bit of a Funky setup, but if that is what was decided . . .

jordan-keenan (approved):
> @jordan-keenan Because of the discrepancy between the default of 2 and how the program is currently coded: The CFG "DEFAULT" value is 2, but the instructions state that a blank value (which would imply the default value) results in no limit. However, the code treated a blank value as the default of 2. The decision was made to leave the CFG Letter file parameter setting for 2 as a default in place, but set the code up such that a blank value in the CFG Letter file would indeed result in there being no limit.

OK. Thanks for explaining, @tkainz . It's a bit of a Funky setup, but if that is what was decided . . .  Approved

rrobison44 (approved):
Looks good.

tkainz:
BA Testing completed - merging program updates

PR #208, opened 2024-10-11
Update BANNO.CDRENEW.V1.POW

tkainz:
Correcting a JSON error when the renewing share term period was greater than 999. When over 999, the JSON output included a comma thousands separator which caused the JSON to error out. Corrective action was to reformat the JSON output as a character value to include the comma for proper display.


jordan-keenan (approved):
Hey @tkainz , The changes look good!

jucarson (approved):
Looks good

PR #209, opened 2024-10-16
BANNO.NEWSUBCREATE.V1.POW fee/funding alpha ID Fix

tkainz:
Removed the value references to the funding source and the fee source share IDs. The existing "VALUE" statements are equating alphanumeric Share IDs to the incorrect ID - usually share ID '0000'


jordan-keenan (approved):
Code changes look good @tkainz !
Am I correct an understanding that in order to allow alphanumeric ID's,
the code now trusts and takes as is the SHARE IDs for this that are sent by the UX?

jucarson (approved):
Code change looks good. Don't forget, when we update something on this one, we should do the same to the custom version in custom-powerons/Custom Sub Account - Business & Consumer.

PR #211, opened 2024-12-11
Banno M2M Transfer

tkainz:
Should FM history have been purged and the program is unable to determine the creation date of the share transfer record, an array out of bounds condition was occurring due to the assigned variable being used for the array index having been initialized to -1.  Program was updated to initialize the TRANSFERCREATEUSER variable to 9999 instead of -1


jucarson (approved):
Change looks good!

jordan-keenan (approved):
Hi Tom, The code changes look good!

PR #212, opened 2025-01-09
BANNO.CDRENEW.V1.POW Update to v1.5.5

tkainz:
1. Update array references on line 2435 and 2530 (share transfers type 2 and 3 indexing) to use TMPLOOP.

2. Modify the code that sets the expiration date on the share note record to use the share’s maturity date.

3. The EXPIREORDELETETYPE3XFERS procedure should not be called for option 1 (increase and renew) on line 478.  We do not want to remove any existing transfers that exist to do the increase.


jordan-keenan (approved):
Code Changes look good Tom! 
Just change the following at the top to reflect 2025 "Copyright 2020-2024 Jack Henry and Associates" please.

PR #216, opened 2025-02-21
error 515 if no shares or loans

rrobison44:
Updated open subshare poweron to prevent the opening of a new share when no shares or loans exist under the member account.  Error 515 is returned in this scenario.

[Jira CUS-3698](https://banno-jha.atlassian.net/browse/CUS-3698)

### DIFF
--- a/open-subshare/REPWRITERSPECS/BANNO.NEWSUBCREATE.V1.CONFIG
+++ b/open-subshare/REPWRITERSPECS/BANNO.NEWSUBCREATE.V1.CONFIG
@@ -54,6 +54,9 @@
 **              Added program installation date parameter setting.
 **  Ver. 1.4.1  09/06/24: JuCarson
 **              Updated help text verbiage.
+**  Ver. 1.5.0  02/26/25 RRobison
+**              Add support for error code 515 in support of: 
+**                Display error if there are no shares or loans under the member account
 **
 **  This Banno service PowerOn is to be installed for use on
 **  demand and is run from the Account Manager workspace to
@@ -151,7 +154,7 @@ DEFINE
  TOCOPENRESULTS         = NUMBER ARRAY(19)
  TOCFILECREATED         = NUMBER
  GROUPLOOP              = NUMBER
- MAXERRORCODE           = 14
+ MAXERRORCODE           = 15
  TESTERRORCODE          = NUMBER
  CONFIGPROGRAMVER       = CHARACTER
 [* TEK 03/30/23]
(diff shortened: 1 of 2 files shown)


tkainz (commented):
The coding changes themselves look fine however, because this is a baseline update, I think that reusing error code 506 is misleading. Error code 506 indicates that there are no shares available to fund a new share opening, when funding is required or that there are no shares available to debit for the opening fee. If funding isn't required and the share can be opened with a $0 balance, this error code is then misleading 

If you do want to continue to use error code 506, then I'd suggest updating the ERRORHANDLER routine to add some error detail info to the output so anyone researching why the member's attempt failed, would have an easier time finding what the true reason was(is).

tkainz (commented):
Line 3469: consider updating the line to indicate all the errors error code 506 applies to.

```
** 506 - No Shares with sufficient funds to xfer OR
         No Shares/Lones with sufficient balance cover Share opening fee(s) OR
         No existing open Share(s) or Loan(s) on the account.
```

tkainz:
> > The coding changes themselves look fine however, because this is a baseline update, I think that reusing error code 506 is misleading. Error code 506 indicates that there are no shares available to fund a new share opening, when funding is required or that there are no shares available to debit for the opening fee. If funding isn't required and the share can be opened with a $0 balance, this error code is then misleading
> > If you do want to continue to use error code 506, then I'd suggest updating the ERRORHANDLER routine to add some error detail info to the output so anyone researching why the member's attempt failed, would have an easier time finding what the true reason was(is).
> > Maybe...
> > ```
> >    ELSE IF ERRORCODE=506 AND ONESHAREFOUND=FALSE THEN
> >     ERRORDETAIL=ERRORDETAIL+" - "+"No existing open Shares/Loans on the account"
> > ```
> 
> Hey Tom. I've asked Emily to review your comments. These were just some changes that she and Tiffany discussed. Maybe we should just add a new error code??

From my perspective, the idea behind error codes is to help make it easier to track down issues, But if we stuff too may issues into one code it kinda defeats that purpose. From the member's perspective, it wouldn't really matter much. So, if at the least, in the JSON code itself, we could add an additional note to the internal error message to indicate whether the 506 error was because of no shares/loans, no fee shares/loans or no opening balance shares/loans would make a big diff when it comes to debugging for us or for providing the CU with more detailed info to better service their members.

rrobison44:
Code was updated to output new error 515 for this new eligibility check logic.  No longer re-using error 506.

tkainz (approved):
Looks good to me - thanks Robby

jordan-keenan (approved):
Code changes look good, @rrobison44 !!

PR #236, opened 2026-04-15
CUS-4500-BANNO.ODTOPTIN.V2.POW-nickname-check

tkainz:
added check for the share nickname - to be used before the share description.


jucarson (approved):
Looks good here too

jordan-keenan (approved):
Changes look good @tkainz

PR #241, opened 2026-07-30
Spanish label fix

rrobison44:
`LBLMN_SP` and `LBLID_SP` did not correctly populate the json labels object so those could not be customized in Spanish.


tkainz (approved):
Thanks Robby

PR #244, opened 2026-09-09
Fix default memo mode message for check withdrawal v2 -- invalid character

rrobison44:
An invalid character made its way into the code when I added the default verbiage during peer review


tkainz (approved):
Looks good Robby - thanks!

PR #246, opened 2026-09-14
CUS-4998_BANNO.CHANGE.ADDR.V1.POW_V2_add_back_hypen_to_9_DIG_ZIP

jordan-keenan:
Added back to 9-digit zip code, the hyphen that gets removed before sent to the RG Var, So the FM works in Symitar. [Jira CUS-4998]( https://banno-jha.atlassian.net/browse/CUS-4998)


change-address/V2/REPWRITERSPECS/BANNO.CHANGE.ADDR.V1.POW
```diff
@@ -3744,4 +3763,22 @@ PROCEDURE BNOPARSEDATALINE
...
 END [PROCEDURE]
 
+PROCEDURE ZIPHYPHEN9DIGADJ
+[* CUS-4998, If RGNEWZIPCODE is 9 long and all numbers,
+** add "-" back at position 6 as it was removed
+** PASS    RGNEWZIPCODE  Updated zipcode
+**
+** RETURN  RGNEWZIPCODE  If 9 long Updated zipcode with hyphen
+*]
+ TMPCHR=RGNEWZIPCODE
```
tkainz: @jordan-keenan I'm not entirely sure what the purpose is of taking the value of TMPCHR and reformatting it as "########9". Is the purpose to remove any non-numeric values? Wouldn't any zip codes that start with a "0" cause issues?  Ie: a legitimate zip code of, say, "012345678" which should equate to "01234-5678" would fail because the formatted value of "012345678" would equate to "_12345678" and then after removing the leading space would leave you with only an 8-digit zip. Since the length wouldn't be 9, then adding it the hyphen wouldn't get completed. Or am I not looking at your intent correctly?

tkainz (changes requested):
Hi @jordan-keenan - can you look at your process for adding in the hyphen - am I looking at it wrong?

change-address/V2/REPWRITERSPECS/BANNO.CHANGE.ADDR.V1.POW
jordan-keenan: Hello @tkainz , Good catch. Yes. I've corrected that in my new commit so now states that have zip codes starting with 0 also work. Instead now I if zip 9 characters I get the value of it as number then turn it to char with a mask that brings back the leading zeros, and if the original number an the new number are equal then I add the hypen.
 IF LENGTH(RGNEWZIPCODE)=9 AND 
    FORMAT("999999999", VALUE(RGNEWZIPCODE))=RGNEWZIPCODE THEN
  RGNEWZIPCODE=SEGMENT(RGNEWZIPCODE,1,5)+"-"+SEGMENT(RGNEWZIPCODE,6,9)

Does the code look good now? I've aready tested that the change works.

change-address/V2/REPWRITERSPECS/BANNO.CHANGE.ADDR.V1.POW
```diff
@@ -106,6 +107,7 @@ DEFINE
   BNOERRMODENTRYLIMIT      = "Address entry exceeds Episys limits"
   BNOERRMODMISSINGDATA     = "Missing data in required field"
   BNOERRMODPOBOX           = "Attempt to enter PO Box as street address or extra address"
+  BNOERRMODZIPLENGTH       = "RGNEWZIPCODE must be 5 or 9 digits"
```
rrobison44: I would remove the RGNEWZIPCODE part of the message.  I think just do: "Zipcode must be 5 or 9 digits".  We don't need to include the name of the internal variable in the error text.

change-address/V2/REPWRITERSPECS/BANNO.CHANGE.ADDR.V1.POW
```diff
@@ -714,6 +716,20 @@ PROCEDURE VALIDATERGDATA
       END
     END
 
+   IF BNOERRORCODE=BNOERRCODESUCCESS THEN
```
rrobison44, line 720: This doesn't completely validate that the zip is in an allowed format (like non-digits are allowed and hyphens are allowed in positions other than pos 6).  Maybe we create and call a procedure like this:
```
PROCEDURE VALIDATEZIP
[ Valide zipcode lengths are 5, 9 and 10 (when contains hyphen in pos 6.  Procedure will return either
  a 5 or 9-digit zip in RGNEWZIPCODE.
]

 ZIPVALID=FALSE
 IF LENGTH(RGNEWZIPCODE)=5 OR
    LENGTH(RGNEWZIPCODE)=9 THEN
  ZIPVALID=TRUE
 ELSE IF LENGTH(RGNEWZIPCODE)=10 AND
         SEGEMNT(RGNEWZIPCODE,6,6)="-" THEN
  DO       
   ZIPVALID=TRUE
   RGNEWZIPCODE=SEGMENT(RGNEWZIPCODE,1,5)+SEGMENT(RGNEWZIPCODE,7,10)
  END 

 [Now make sure we only have digits in the zip]
 IF ZIPVALID=TRUE
  DO
   FOR TMPLOOP=1 TO LENGTH(RGNEWZIPCODE)
    DO
     IF CHARACTERSEARCH("0123456789",SEGMENT(RGNEWZIPCODE,TMPLOOP,TMPLOOP))=0 THEN
      ZIPVALID=FALSE
    END  
  END
END
```

change-address/V2/REPWRITERSPECS/BANNO.CHANGE.ADDR.V1.POW
```diff
@@ -3744,4 +3763,18 @@ PROCEDURE BNOPARSEDATALINE
  BNODLREMOVEPADDING=FALSE 
 END [PROCEDURE]
```
rrobison44, line 3767: If you implement the VALIDATEZIP function as mentioned above, this line can be removed:
```
FORMAT("999999999", VALUE(RGNEWZIPCODE))=RGNEWZIPCODE
```
This code is sorta combining adding back the dash but also doing some vague data validation when you check RGNEWZIPCODE against the "999999999" formatted value of RGNEWZIPCODE.

rrobison44 (changes requested):
See comments.

change-address/V2/REPWRITERSPECS/BANNO.CHANGE.ADDR.V1.POW
jordan-keenan: Thanks @rrobison44 ; I've made that change.

change-address/V2/REPWRITERSPECS/BANNO.CHANGE.ADDR.V1.POW
jordan-keenan: Thanks @rrobison44 , Now checking that the 10 digit also has the hyphen in it following where the code was headed so we can strip out the hyphen (to be consistent - right now Banno doesn't send it to us) and add it back during FM. So yes I've added those good changes and they are working.

change-address/V2/REPWRITERSPECS/BANNO.CHANGE.ADDR.V1.POW
jordan-keenan: Thanks @rrobison44 , I preferred to do that (already have the hyphen out if we are adding it back later) during validation as well; Glad that is back so most is done in one step. And the final step of adding the hyphen back before FM makes sense (though could have been done in the verify step, but is more organized by doing it before FM).

rrobison44 (approved):
Changes look good.

tkainz (approved):
Looks much better Jordan - thanks. (Approval assumes Robby's JSON contract update is added

jucarson (approved):
Changes look good @jordan-keenan!

